PR #1, opened 2017-06-29
Add Vi Mode

apm1467:
The Vi Mode I documented here: http://yige.ch/vi-mode-for-karabiner-elements/

This was the Vi Mode on the old Karabiner. Now I've implemented it for the new Karabiner Elements.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -69,6 +69,18 @@
       </div>
     </div>
 
+          <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Vi Mode</h3>
+      </div>
+      <div class="panel-body">
+        <ul>
+          <li>Change S to Left_control + Fn if pressed with other keys (Post S when pressed alone)</li><li>Change Left_control + Fn + H/J/K/L to Arrow Keys</li><li>Change Left_control + Fn + F to Fn</li><li>Change Fn + H/J/K/L to Home / Page Down / Page Up / End</li>
+        </ul>
+        <a class="btn btn-primary btn-sm" style="margin-left: 40px" href="karabiner://karabiner/assets/complex_modifications/import?url=https%3A%2F%2Fpqrs.org%2Fosx%2Fkarabiner%2Fcomplex_modifications%2Fjson%2Fvi_mode.json">Import</a>
+      </div>
+    </div>
+
     </div>
 
     <script src="assets/jquery-3.2.1.min.js"></script>
--- a/docs/json/vi_mode.json
+++ b/docs/json/vi_mode.json
@@ -1,0 +1,138 @@
+{
+    "title": "Vi Mode",
+    "rules": [
+        {
+            "description": "Change S to Left_control + Fn if pressed with other keys (Post S when pressed alone)",
+            "manipulators": [
+                {
+                    "type": "basic",
+                    "from": {
+                        "key_code": "s",
+                        "modifiers": {
+                            "optional": ["any"]
+                        }
+                    },
+                    "to": [{
+                        "key_code": "fn",
+                        "modifiers": ["left_control"]
+                    }],
+                    "to_if_alone": [{"key_code": "s"}]
+                }
+            ]
+        },
+        {
+            "description": "Change Left_control + Fn + H/J/K/L to Arrow Keys",
+            "manipulators": [
+                {
+                    "type": "basic",
+                    "from": {
+                        "key_code": "h",
+                        "modifiers": {
+                            "mandatory": ["left_control", "fn"],
+                            "optional": ["any"]
+                        }
+                    },
+                    "to": [{"key_code": "left_arrow"}]
+                },
+                {
+                    "type": "basic",
+                    "from": {
+                        "key_code": "j",
+                        "modifiers": {
+                            "mandatory": ["left_control", "fn"],
+                            "optional": ["any"]
+                        }
+                    },
+                    "to": [{"key_code": "down_arrow"}]
+                },
+                {
+                    "type": "basic",
+                    "from": {
+                        "key_code": "k",
+                        "modifiers": {
+                            "mandatory": ["left_control", "fn"],
+                            "optional": ["any"]
+                        }
+                    },
+                    "to": [{"key_code": "up_arrow"}]
+                },
+                {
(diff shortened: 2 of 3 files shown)


tekezo:
Thanks!

Liquidmantis:
This is really awesome, but it makes typing problematic at speed.  Shift-S won't register a capital S without deliberately slowing, and S's at speed either don't get entered or will cause unexpected behavior.  I tried adjusting the to_if_alone_timeout but couldn't work around the issue.  Unfortunately, I can't leave it enabled as is.

apm1467:
I've meet the same problem. 

The `to_if_alone_timeout` isn't helping here. In the old Karabiner, there was a `__BlockUntilKeyUP__` timeout parameter. That was the important timeout parameter for Vi Mode. It hasn't been implemented in Karabiner Elements yet, but I think it is coming. Meanwhile, you may consider using the built in "right_command + H/J/K/L" rule. 

I'm also trying to improve the Vi Mode. Currently it uses `Ctrl-Fn` as a fake modifier key. If S is pressed with other keys, a `Ctrl-Fn-otherKey` event will be posted. `Ctrl-Fn-H/J/K/L` are remapped to arrow keys, but `Ctrl-Fn-A` for example, will be treaded as `Ctrl-A` by the OS. This is the reason for the unexpected behaviors.

apm1467:
@Liquidmantis I've fixed all the problems you mentioned. 

It turns out they are mostly not Karabiner-Elements' problems, but my problems of implementation. Sorry about that. Now you can give Vi Mode another try.

Liquidmantis:
@apm1467 Nice!  Thanks for working on it.  It seems much better now; however, there's still a subtle issue where it seems like there's a "hang time" with the s key.  I'm seeing that typing sh, sk, and sl words (e.g. shore, sky, slow) will instead do the arrow movement.  

In testing I'm able to reproduce it with distinctly sequential, albeit rapid, key presses.  In fact, while typing the prior sentence I didn't get the 's.' on the end of presses for the same reason.

apm1467:
@Liquidmantis Improved now. Please update again, the 's.' case should not happen now.

The "sh, sk, and sl words" is however not a problem I can solve.

If we first press S, than release it completely, then press H/J/K/L, they will work fine. 

But when we type fast, what we sometimes do is:

1. Press S
2. Press another key
3. Release S
4. Release another key

The "Press another key" event happens before "Release S" event. 

The old Karabiner did handle it very well [here](https://github.com/tekezo/Karabiner/blob/05ca98733f3e3501e0679814c3795d1cb57e177f/src/core/kext/Classes/EventInputQueue.cpp#L837). The key events order will be changed in input event queue. I can only hope @tekezo also implement this in Karabiner-Elements.

Liquidmantis:
Thanks, @apm1467!  The 's.' case is definitely handled now.  

Before my prior report I did wonder if it was a simple j/k/l/; keyDown event before the s keyUp.  So, I tried just using a "flam" [druming] technique to test and in so much as I'm able to test it is occurring with those well-timed discrete presses.  I tested with my MacBook's internal keyboard as well as my Apple 2nd gen Magic Keyboard.  It could just be that these chiclet keys have enough latency at this millisecond timing that the keyUp event lags behind the key release.  

I was testing with the phrase "She sells sea shells by the sea shore", which is a contrived test case.  I'll continue trying it in more real world usage and see how it works while hoping for better internal handling as KE matures.

Liquidmantis:
Unfortunately, it didn't last long for me.  First thing this morning when I was back at it I was running into backspacing errors (typing "aren't showing" in an email resulting in aren'towing).

Have you thought about combination activator like the s+f one?  Using 'as' as the activator for for j/k/l/; should eliminate the issue.  If I get a chance today I might take a crack at setting that up.

apm1467:
This might be the case that you typed "aren't sh", and "sh" is trigged as left arrow, so the text cursor goes one position back (before the space). And then you continue typing "owing", so the result is actually "aren'towing " (with a space at the end).

This is then not a backspacing error, but again a sh-word error. 

I've thought about using D instead of S as the activator, since there are not so many dh-, dj-, dk- and dl-words. The ultimate solution is still wait for KE to implement input event queue support. I might wait for a little bit time now; if that support still does not come, I will change from letter S to letter D.

apm1467:
So I've done that anyway. Now you can give the D key a try. It does produce far less typing errors. Just my muscle memory is still on the S key.

wincent:
I filed an issue for the `__BlockUntilKeyUp__` functionality: https://github.com/tekezo/Karabiner-Elements/issues/877

kamykaze:
@apm1467 I was thinking about this, and while I don't exactly use your vi mode (i prefer the Cmd+h/j/k/l mapping better), I thought of a possible workaround for the time being.

What if you do the following:

- S to trigger 'vi_mode' on
- HJKL to arrow mappings if 'vi_mode' is on 
- 'any' key to trigger 'vi_mode' off, send key code

Basically, when you're in vi_mode (holding down S), hjkl will trigger arrow keys, anything else will move down to the next rule, which will turn vi_mode off. 

The problem is that I don't know if there's a way to catch "any" key, and remap it with the same key. I know there's a way to capture any key, but I think we just lose that info.

An alternative is to map every key on the keyboard to itself except the keys you want special mapping (see my tmux prefix mode that changes Fn + any key -> Ctrl+Space, any key)

apm1467:
@kamykaze This is pretty much the current behaviour of Vi Mode. 

The problem of Vi Mode is that sometimes when we type S + H fast, what we want is the letters "sh", not a left_arrow.

apm1467:
There's now a new Simple Vi Mode in beta version available. You can test it. It may solve the problem of the current Vi Mode. 

https://pqrs.org/osx/karabiner/complex_modifications/#personal_tekezo_simple_vi_mode
https://github.com/pqrs-org/KE-complex_modifications/blob/9c60286418c353d68fec5b2729a7fe510a7cd865/docs/json/personal_tekezo_simple_vi_mode.json

apm1467:
The [Vi Mode (rev 5)](https://pqrs.org/osx/karabiner/complex_modifications/#vi_mode) is available. Now the S key can be used as trigger key perfectly. It requires the newest version of Karabiner-Elements (12.1.0 or later).

PR #4, opened 2017-06-30
Added configuration for producing f-keys by tapping modifiers-keys.

runartrollet:
This configuration will produce f-keys by tapping modifier-keys. For instance, pressing `left_command` produces `F14` and so on.


tekezo:
Thanks!

PR #15, opened 2017-07-09
Create "change control and Fn + letter keys to command and control keys" complex modificator

marcoamarelo:
Create a correspondent erb file for this modificator is still necessary.


tekezo:
Thanks!

PR #16, opened 2017-07-10
Added configuration for exchanging left command + arrows and left option + arrow

matijasss:
Useful for speed code selection and navigation, allows to move between, select and copy words without moving left finger from command key. It can be also useful for new Mac users coming from Windows.


tekezo:
Thanks!

PR #19, opened 2017-07-11
update Vi Mode (rev 2.1)

apm1467:
improve punctuation mark and number keys handling


apm1467:
improve punctuation mark and number keys handling in Vi Mode

PR #25, opened 2017-07-12
Create rules for blocking one-handed key combo

briggySmalls:
Creates rules to ignore key combinations (standard key + modifier) that are performed using one hand.

**Rationale**: I have been developing strain in my wrists while typing, and I think that it is partly due to contorting my hand into uncomfortable positions attempting to hit combinations such as _⌘ + a_ or _⇧ + ]_ with one hand.

The 8 rules introduced are:
- Left ⌘ + key
- Left ⇧ + key
- Left ⌥ + key
- Left ctrl + key
- Right ⌘ + key
- Right ⇧ + key
- Right ⌥ + key
- Right ctrl + key

Where 'key' is a corresponding alpha-numeric, or other symbol key that is on the same side of the keyboard as the modifier, as it is laid out on my macbook. All of the rules permit any number of optional, additional modifiers.

Thanks! 😁


tekezo:
Thanks!

PR #26, opened 2017-07-13
Added binding rule "Command_R + Command_L to escape"

ymkjp:
#### Summary
I've added new rule which originally came from Karabiner

#### Verification
* [x] Local testing

Let me know if any concerns.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -177,6 +177,18 @@
       </div>
     </div>
 
+          <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Change Command_L Key (Left Command)</h3>
+      </div>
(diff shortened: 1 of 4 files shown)


tekezo:
Thanks!

ymkjp:
#### Trouble shooting
In case _"Command_R + Command_L to escape"_ doesn't work for you, make sure to prioritize this rule over others which hook `right_command` or `left_command`.

Screenshot_2017-07-17_15_05_38

Verified this rule conflict occurs with [japanese.json](https://github.com/pqrs-org/KE-complex_modifications/blob/1b3740d72072c7a3c41e2055c205b61fe2ad5f76/docs/json/japanese.json) at Karabiner-Elements _v0.91.6_.

PR #28, opened 2017-07-13
added virtual numeric keypad using Tab

kamykaze:
Added a virtual numeric keypad by holding Tab. This is useful if you are typing numbers and don't have a full keyboard.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -221,6 +221,18 @@
       </div>
     </div>
 
+          <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Numeric Keypad</h3>
+      </div>
+      <div class="panel-body">
+        <ul>
+          <li>Use a virtual numeric keypad by holding Tab. Pressing Tab alone sends Tab key</li>
+        </ul>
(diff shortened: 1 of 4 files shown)


apm1467:
This may conflict with the Vi Mode. Control + Fn + J/K/L are defined as arrow keys in Vi Mode, but number keys in this rule. So they will basically overwrite each other.

Would you consider using something like Option + Fn instead of Control + Fn as modifier keys in this rule?

tekezo:
Thanks!

kamykaze:
@apm1467 You're right. I copied the vi mode file to start this, which is why the modifier keys are the same. I don't care for the modifier keys Option + Fn because ultimately I'm using Tab. But it seems a combination of modifier is needed in order to make Tab work as Tab if nothing else is pressed. 

I'll have to do more testing to see if I can achieve the same without any Ctrl/Opt/Fn modifiers, but I'll change it to Control + Fn if that doesn't work.

apm1467:
Yes, the Vi Mode is kind of a hack. Since the S key (or in this rule, the Tab key) is not a modifier key, it can not build a key combination with another non-modifier key.

So the Vi Mode changes the S key into a "fake" modifier Control + Fn, which is a modifier that almost no one is using, so it can build valid key combinations with other keys, and won't cause conflict with normal modifier keys usage.

What your rule could do, is to choose another rarely used "fake" modifier, so it will not cause conflict with normal usage. 

I do want to see if this can be achieved without such kind of hacking. Good luck with testing.

PR #31, opened 2017-07-16
Enable HHKB compatible mode

tockrock:
This adds some of the HHKB compatible Arrow Key Mode that was available in Karabiner, with an extra media keys

https://github.com/tekezo/Karabiner/blob/05ca98733f3e3501e0679814c3795d1cb57e177f/src/core/server/Resources/include/checkbox/device_specific.xml#L518-L525

This is for the US layout (no English documentation available?)
http://www.pfu.fujitsu.com/hhkeyboard/manual/P3PC-5931-01.pdf

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -233,6 +233,18 @@
       </div>
     </div>
 
+          <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Happy Hacking Keyboard Compatible Mode (rev 1)</h3>
+      </div>
+      <div class="panel-body">
+        <ul>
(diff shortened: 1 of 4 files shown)


tekezo:
Thanks!

tockrock:
Thank you for developing this tool!
It's really essential to me, and I'm sure many others as well!

apm1467:
There is an undocumented key combination on HHKB that I use all the time:

Fn + O/P -> change display brightness

Many people might not know this :)

tockrock:
Ah cool, thanks for the heads up!
Found the QA.
https://github.com/pqrs-org/KE-complex_modifications/pull/31#issuecomment-315651451

PR #32, opened 2017-07-16
Use right_command+h/j/k/l/;/' to switch current tab/window/application

kzsh:
Use `⌘+;/'` switch applications (instead of ⌘+Tab,) `⌘+j/k` switch open windows and `⌘+h/l` switch tabs.  It is intended to offload some common work from the left hand to the right hand in order to better distribute physical effort.

### DIFF
--- a/docs/json/application_window_tab_switch_with_right_cmd_hjklsemiquote.json
+++ b/docs/json/application_window_tab_switch_with_right_cmd_hjklsemiquote.json
@@ -1,0 +1,175 @@
+{
+  "title": "Use right_command+h/j/k/l/;/' to switch current tab/window/application",
+  "rules": [
+    {
+      "description": "Use right_command+;/' to cycle through running applications (like command+tab).",
+      "manipulators": [
+        {
+          "type": "basic",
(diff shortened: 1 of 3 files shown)


runartrollet:
Isn't this something that is more easily changed within the System-config in MacOS?

kzsh:
@runartrollet Interesting!  It hadn't occurred to me to look in the standard keyboard shortcuts.  When you say System-config, do you mean System Preferences -> Keyboard, or something lower level?  I'm not familiar with lower-level options.

If it's with System Preferences -> Keyboard, I see that you can remap "Move focus to next window" which would accomplish part of my goal, but would require me to type, for example, `⌘+j` `⌘+shift+j` to cycle backwards rather than using both j and k as a kind of intuitive next-previous (at least for a vi-style layout).

As for the other mappings, I didn't see a way to change them, so I'll wait for your reply to comment.

Thanks for the feedback!

runartrollet:
You are right, you cannot set the keyboard-shortcut for cycling backwards. Thank you for clearing this up.

PR #37, opened 2017-07-17
keybindings to launch new windows of Chrome or iTerm2

smlewis:
As requested here: https://github.com/tekezo/Karabiner-Elements/pull/837, I am also putting this JSON in this repository.  

copied description from other PR:
This JSON example lets you open a new iTerm2 window via "left control, left shift, and `" or a new Chrome window via "left control, left shift, and 1", using the new shell_command in the complex modifications json element. (The ability to launch new windows, and not just applications, via keyboard bindings is much harder on Mac than Linux, and may be useful for recent converts like me trying to salvage their old Linux habits via Karabiner).

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -211,6 +211,19 @@
 
           <div class="panel panel-default">
       <div class="panel-heading">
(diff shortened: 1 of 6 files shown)


tekezo:
Thanks!

PR #38, opened 2017-07-17
Added rules for Microsoft Remote Desktop Client

bnickel:
Rules swap Command and Control keys so key combinations like Command-C
work as expected.


tekezo:
Thanks!

PR #40, opened 2017-07-20
change_command_r

iofik:
Helpful for vim-users having TouchBared macs when other options (like CapsLock) are not good for some reason.


tekezo:
Thanks!

PR #42, opened 2017-07-21
Add shift-LR -> caps_lock

Huluk:
Toggle caps lock by pressing left Shift and right Shift at the same time.
This frees up the caps key for other modifications.


tekezo:
Thanks!

PR #47, opened 2017-07-23
*Added rule*: Caps and enter to ctrl. Ctrl to caps.

grhawk:
This idea of a complex-rules repo is very very nice!  Wonderful Job. Below the commit string.

I find this rule extremely useful when using emacs!


tekezo:
Thanks!

PR #54, opened 2017-07-25
[JIS] "Shift+2" to "@"

spookyway:
Title says all

### DIFF
--- a/src/json/jis-at-symbol
+++ b/src/json/jis-at-symbol
@@ -1,0 +1,19 @@
+{
+   "title": "JIS-to-ISO-for-Macbook-Pro-testing"
+   "rules": [
+     {
+       "description": "Change Shift + 2 to at symbol",
+       "manipulators": [
+       {
+         "type":"basic",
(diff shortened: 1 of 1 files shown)


spookyway:
Thank you for your feedback. Later I'd update this project by changing other JIS elements to corresponding ISO ones as far as possible.

tekezo:
Does not this configuration help you?

spookyway:
It does, but partly. "¥" and "_" keys do nothing under the virtual keyboard layout. Some keys are missing as well. Now I'm using "kana" key as tilda . . . but the layout can be improved in more effective way.

tekezo:
OK, please fill the keys which is not possible by keyboard type.

Note:
I recommend you to use simple_modifications if keys can be simply exchanged to avoid conflict another complex_modifications rules.
Karabiner-Elements applies only one complex_modifications for one event.
If you use complex_modifications to layout changes, other complex_modifications rules (e.g., emacs key bindings) will be ignored for the keys.

tekezo:
I close this PR for now.
Please make new PR when you made a complete definition.

PR #57, opened 2017-07-26
Added rule - ALT<=>CMD in Autodesk Maya

kurtontheway:
Added a rule of swapping option and command in Autodesk Maya.


tekezo:
Thanks!

PR #61, opened 2017-07-28
Added tmux prefix bindings and each_key() to remap multiple keys at once

kamykaze:
I created a tmux prefix mapping to basically handle the following remappings:

Fn+key --> <Ctrl+B>, key
(eg: Fn+c becomes Ctrl+B, c)

Where key is anything from a-z, 0-9, and some other keys.

Obviously, copying similar code multiple times is not very maintainable, so I wrote each_key() inside erb2json.rb to handle situations like this.

Please take a look and let me know if I wrote that properly. I'm not a Ruby programmer, but I think it works pretty well.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -450,6 +450,36 @@
       </div>
     </div>
 
+          <div class="panel panel-default">
+      <div class="panel-heading">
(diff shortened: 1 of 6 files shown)


tekezo:
This is awesome!
Please merge it by yourself.

PR #63, opened 2017-07-28
Dvorak Keyboard layout; updated each_key() to have two sets of keys

kamykaze:
In order to create a quick way to map a lot of keys from one key to another, I updated each_key() to include one more parameter: dest_keys_list. Now you can easily map "a", "b", "c" to "x","y","z", which is very useful for creating recipes for alternative keyboard layouts like Dvorak. (Note: I had to updated my tmux recipe to work with this)

In addition, I renamed from() and to() to _from() and _to() and made it only return the data object (instead of printing the JSON generated).  This allows more flexibility so I can use those functions inside each_key() without having to parse the JSON. 

But don't worry, I added wrapper from() and to() so all existing recipes are still valid.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -505,6 +505,20 @@
     </div>
 
 
+      <!-- Full keyboard layouts -->
+          <div class="panel panel-default">
+      <div class="panel-heading">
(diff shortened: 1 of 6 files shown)


tekezo:
Thanks!

PR #66, opened 2017-07-29
Map fn+numbers to corresponding function keys

bingocaller:
This solves [#847](https://github.com/tekezo/Karabiner-Elements/issues/847) in a pretty straightforward manner.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -516,6 +516,18 @@
       </div>
     </div>
 
+          <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Map fn + number keys to function keys</h3>
+      </div>
+      <div class="panel-body">
+        <ul>
+          <li>Map fn + number keys to their corresponding function keys</li>
+        </ul>
+        <a class="btn btn-primary btn-sm" style="margin-left: 40px" data-json-path="json/fn_plus_numbers_to_function_keys.json">Import</a>
+      </div>
+    </div>
+
 
       <!-- Full keyboard layouts -->
           <div class="panel panel-default">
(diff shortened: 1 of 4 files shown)


tekezo:
Thanks!

bingocaller:
@tekezo No problem! This should work, right?
[It's not working for @dluxhu as is](https://github.com/tekezo/Karabiner-Elements/issues/847), but I can't figure out why? Right now I can't get _any_ Modifications to work on my machine; neither simple nor complex ones.

kamykaze:
I tested this. It mostly works. One thing I had to make sure is to check the "Use all F1, F2, etc. keys as standard function keys" option under the Function keys tab of Karabiner. Otherwise you get those hardware keys for Display brightness, etc.

There is one bug though: F11 and F12 don't work. This is because you have "-" and "=" as the key code. It should be "hyphen" and "equal_sign"

bingocaller:
@kamykaze Haha, whoopsie! 😅
I actually have another commit in the pipeline which uses the `_each_key` method to generate the JSON for all the remappings; now I can get the correct key names in there. Maybe also a note in the description regarding "Use all F1, F2…"?
Anyways, thanks for testing this for me. I still haven't been able to make Modifications work on my machine. 😕

kamykaze:
yeah, those changes would be nice. I hope you get your mappings working. Any idea what's wrong? Maybe create an issue here in KE repo?

bingocaller:
I don't have a clue what's wrong with my mappings. I've tried a few things already, but I think I'll create an issue to track it there.

bingocaller:
Hmm.. I just updated to v0.91.12 and now everything is working fine! 😌

PR #72, opened 2017-07-31
Add Mu Mode Semicolon(;)+J/K/L/I/U/M/O/. to Arrow Left/Down/Right/Up/PageUp/PageDown/Home/End

muyinliu:
Mu Mode Semicolon(;)+J/K/L/I/U/M/O/. to Arrow Left/Down/Right/Up/PageUp/PageDown/Home/End

Note: with only right hand also can press Arrow keys(on HHKB it's great!).

### DIFF
--- a/docs/json/mu_mode.json
+++ b/docs/json/mu_mode.json
@@ -1,0 +1,226 @@
+{
+  "title": "Mu Mode (rev 0.1)",
+  "rules": [
+    {
+      "description": "Mu Mode Semicolon(;)+J/K/L/I/U/M/O/. to Arrow Left/Down/Right/Up/PageUp/PageDown/Home/End",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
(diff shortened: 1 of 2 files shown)


kamykaze:
Did you write the json file from scratch? I think preferably we want the file to be generated from its .erb file. That makes the original file simpler and easier to maintain.

smlewis:
@kamykaze Just to note - the README suggests that writing the JSON directly into docs/json is ok (first line under 'Add Rules') - if it's not OK then perhaps the README should say so?  (I don't know what the maintainers want the rule to be, I'm just commenting on the discrepancy, as a mostly-clueless user)

tekezo:
Thanks!

muyinliu:
@tekezo How long will this config show on page [Karabiner-Elements complex_modifications rules](https://pqrs.org/osx/karabiner/complex_modifications/) after merged?

muyinliu:
@kamykaze thanks.

PR #73, opened 2017-07-31
Rubocoptero personal rules related to compose key

rubocoptero:
I've a keyboard layout that makes ctrl+shift+f13 behave as a compose key so I've added a couple of personal rules according to it.


tekezo:
Thanks!

PR #78, opened 2017-08-01
Update each_key with a few niceties

bingocaller:
This PR combines `each_key` and `_each_key` into one function, which leverages keyword arguments to make function calls more readable and less verbose (i.e. you don't have to pass unused arguments). It also adds a check for presence of `conditions` to prevent them from showing up as a value-less key in the final JSON file.

### DIFF
--- a/docs/json/dvorak_layout.json
+++ b/docs/json/dvorak_layout.json
@@ -24,9 +24,6 @@
             {
               "key_code": "grave_accent_and_tilde"
             }
-          ],
-          "conditions": [
-
           ]
         },
         {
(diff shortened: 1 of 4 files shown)


kamykaze:
Great improvements. I should have thought to use keywords arguments when the list of arguments grew.

I think you just missed a few instances of _each_key() or it was added after. Once those are fixed we should merge this.

PS: Not sure if you're aware, but run `make rebuild` to force all json files to recompile. I ran into a similar issue before where `make all` only recompiles .json.erb files that were changed.

PR #83, opened 2017-08-04
Pok3r Escape Key

teonivalois:
* Adding a complex modification to support Pok3r style escape key. [_grave_accent_and_tilde_] becomes [_escape_] and [_fn_]+[_grave_accent_and_tilde_] becomes [_grave_accent_and_tilde_].
* No changes to the current behavior of the [_escape_] key are made.
* Pairs really well with the "Map fn + number keys to function keys" on new MacBooks (>= 2016) where the [_escape_] and function keys [F1 ~ F12] are all virtual keys in the touch bar.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -351,6 +351,15 @@
     </div>
     <div class="panel panel-default">
       <div class="panel-heading">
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #84, opened 2017-08-08
Improve HHKB mode

tockrock:
Improves use of arrow keys with the Shift key, and also adds mapping to F13-15 with fn i, o, p.


muyinliu:
Thanks

PR #92, opened 2017-08-15
Add caps lock to slash/backslash (German keyboard layout)

jomec:
Since it's not possible to address slash and backslash with Karabiner on a German keyboard, I added these two complex mods in order to redirect caps lock to these keys.

### DIFF
--- a/docs/json/caps_lock_german.json
+++ b/docs/json/caps_lock_german.json
@@ -1,0 +1,63 @@
+{
+  "title": "Change caps_lock key (German keyboard layout)",
+  "rules": [
+    {
+      "description": "Change caps_lock key to slash.",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "caps_lock",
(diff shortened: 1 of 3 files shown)


docs/json/caps_lock_german.json
```diff
@@ -0,0 +1,63 @@
...
+            "key_code": "caps_lock",
+            "modifiers": {
+              "optional": [
+                "any"
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "7",
```
muyinliu: Make sure `"key_code": "7"` is correct. It looks like that  `191` is the **key_code** on most keyboard.

jomec: Yes, unfortunately the german keyboard layout has no keys for slash or backslash. They are the alternate functions of the key "7". The script works fine for me.

muyinliu: OK.

PR #95, opened 2017-08-19
Add touchcursor_mode, as a much simplified copy of vi_mode

jeebak:
This is a much simplified copy/port of @apm1467 `vi_mode` to [TouchCursor](https://github.com/martin-stone/touchcursor) mappings.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -24,7 +24,7 @@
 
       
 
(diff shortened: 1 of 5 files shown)


apm1467:
Thanks!

You may face some problems that the Vi Mode is currently facing: 

Sometimes when typing `space + letter` fast, what we want is really a space and a letter, not an arrow or home key or something like that.

Because of this reason, the Vi Mode has avoided using common letter combinations in order to reduce trigger errors. The Touch Cursor Mode is using a lot of common letter combinations like `space + y` and `space + i`, so it might be annoying in daily usage. Let's wait for Karabiner-Elements to get better support for this issue.

jeebak:
Indeed. This problem existed for the TouchCursor mapping in the older (pre-Elements) version of Karabiner as well, but after cost/benefit analysis, having the TC mappings in KE will make the upgrade to macOS Sierra smoother for people like me who's held off because we can't live without it :). Thanks for making that possible by providing the `vi_mode` mappings to use as an example!

PR #96, opened 2017-08-19
Adds JIS to ASCII complex modifier

agektmr:
This patch creates a file with 3 rules which make Japanese JIS keyboard layout behave like US ASCII keyboard layout.


tekezo:
Thanks!

PR #97, opened 2017-08-21
Add complex modifications for SpaceFN

U47:
This adds rules surrounding SpaceFN (see: https://geekhack.org/index.php?topic=51069.0), broken down into several chunks so people can pick and choose what parts they want.


tekezo:
Thanks!

PR #100, opened 2017-08-26
Add PC Style Redo and Correct Ctrl Left+Right

beauallison:
Currently Ctrl+Left/Right moves the insertion point to the beginning/end of the line; this should instead move to the beginning of the previous/next word (Option+Left/Right Arrow).

I've also added a PC Style Redo (Ctrl + Y).


apm1467:
Thanks!

PR #101, opened 2017-08-28
do not fix path to ruby executable (for make)

rcmdnk:
With `/usr/bin/ruby` fixed path,
`make` fails if user installed ruby by such Homebrew and has GEM_HOME for different version.

    $ make
    sh scripts/update-json.sh
    /usr/bin/ruby scripts/erb2html.rb < src/index.html.erb > docs/index.html
    /usr/local/gems/gems/json-2.1.0/lib/json/ext/parser.bundle: [BUG] Segmentation fault
    ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]

    -- Crash Report log information --------------------------------------------
   ...

It is better to use `#/usr/bin/env ruby` as shebang, and commands are not necessary in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 all:
-	sh scripts/update-json.sh
-	/usr/bin/ruby scripts/erb2html.rb < src/index.html.erb > docs/index.html
-	/usr/bin/ruby scripts/erb2html.rb < src/example.html.erb > docs/example.html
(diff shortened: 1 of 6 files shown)


apm1467:
Thanks! I can confirm this works fine on my machine. Can anyone also review this?

tekezo:
Thanks!

PR #103, opened 2017-08-29
Add german PC-Style shortcuts

TcT2k:
A few symbols are quite different from the german PC layout and mac layout, especially curly brackets, pipe, tilde, backslash. This adds mappings for those.


apm1467:
Thanks!

PR #109, opened 2017-09-05
Don't exit fullscreen when pressing ESC in Safari

S3RK:
Pressing ESC to exit fullscreen videos at some popular sites still works


apm1467:
Thanks!

PR #113, opened 2017-09-08
exclude vscode for emacs_key_bindings

yqrashawn:
Emacs key bindings have conflicts with vscode vim extension. 
Extension users may feel the extension is broken when they do not know, for example, this complex modification map <C-d> to delete. 
And user can always customize emacs like or vim like keybindings in vscode through extensions.

So I recommend disable this in vscode by default.


tekezo:
Thanks!

I also updated erb2json.rb by this commit.
(We have to update erb2json.rb because the json file is generated by this script.)

https://github.com/pqrs-org/KE-complex_modifications/commit/40392c57675c59dab59f7a4cf8828449ff90d736#diff-ec216709002c75751c08467fb890337a

PR #116, opened 2017-09-14
Add complex modifications for exchanging control+tab and command+tab

Drazcmd:
This makes using capslock->control instead of left_control on mac
keyboards much nicer for people like me (i.e. those who switch between
tabs more often than between applications).


muyinliu:
Thanks.

PR #128, opened 2017-09-23
change grave accent to escape

adamfeuer:
### Summary

* Adds a new complex modification that makes backtick into Escape, giving a physical Escape key that is in a similar position to the one on older Macs.

### Details

* change Backtick (grave accent) to Escape
* change Option-Backtick into Backtick
* leave Shift-Backtick as Tilde as usual
* leave Command-Backtick so it also works as normal


tekezo:
Thanks!

PR #129, opened 2017-09-24
Swap keybindings of cmd+arrows and ctrl+arrows

marchbnr:
Similar to the existing script to swap "command + arrow keys" and "option + arrow keys", this script swaps between "command + arrow keys" and "control + arrow keys".


apm1467:
Thanks!

PR #130, opened 2017-09-24
Add a template to exchange command+backspace and option+backspace

marchbnr:
A new script to swap between "command + backspace" and "option + backspace", effectively swapping "delete current line" and "delete current word"


apm1467:
Thanks!

PR #135, opened 2017-09-26
Post space is pressed alone, left_ctrl otherwise

mmarkuss:
Handy feature for emacs users. Single space alone posts space. Space + other key = Ctrl + other key. 
There is one problem with delay. Typing fast with spaces is sometimes understood as Ctlr.


apm1467:
Thanks!

Since there is already a "Change spacebar" rule, I will combine your rule with it together.

PR #140, opened 2017-10-06
Disable PC-Style Quit (Alt+F4) for RDC and VMs

TcT2k:
Previous modification did not disable this mapping for RDC and VMs

BTW
make didn't seem to be run after a recent change to include Teamviewer as a simple run of make on the repo added a lot of changes to docs/json/pc_shortcuts.json (which I didn't commit)


apm1467:
Thanks!

I will check if a remake is needed.

PR #141, opened 2017-10-06
adds vi style 3, which maps fn-hjkl to the arrow keys

dwmkerr:
Hi! I've found this mapping enormously helpful, it's just another variation on the modifier key + hjkl vim theme. Thanks for the great software, not sure how I did without it!

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -24,8 +24,7 @@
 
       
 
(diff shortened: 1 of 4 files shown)


apm1467:
Thanks!

PR #146, opened 2017-10-08
Add Neo2 layout.

jgosmann:
Add rules for the [Neo2 keyboard layout](http://neo-layout.org/).

Intended to be used with [the corresponding keyboard layout file](https://github.com/jgosmann/neo2-layout-osx).


tekezo:
Thanks!

PR #150, opened 2017-10-11
Add japanese-specific vim's rule

kosukemori:
hi @tekezo !

This is my suggestion to add the following rules in japanese rule
- "Escape to Escape & Japanese_Eisuu"
- "Ctrl+[ to Ctrl+[ & Japanese_Eisuu"

Basically these rules are for vim. However I think some people may like this rules by different reasons.

Ideally, I don't want add "Ctrl+[ to Ctrl+[ & Japanese_Eisuu" because it's not beautiful. But "Escape to Escape & Japanese_Eisuu" couldn't cover "Ctrl+[", so I added the rule.


tekezo:
Thanks!

Issue #153, opened 2017-10-14
Volume buttons to Opt Shift Volume (fine volume control)?


mrelwood:
I'm not able to make this work. I'm a noob, and there is no precise easy-to-follow guide on how to make my own complex modifications. So I probably did something wrong.

Goal is to have fine Volume controls without needing to press Option and Shift as well. I was surprised there isn't a ready model for this, as I've wanted this for years! Can it be done?

bingocaller:
@mrelwood Try saving this as a JSON-file in your `~/.config/karabiner/assets/complex_modifications` folder:
```json
{
  "title": "Fine volume contol",
  "rules": [
    {
      "description": "Map volume controls to fine volume controls (Option + Shift + Volume)",
      "manipulators": [
        {
          "type": "basic",
          "from": {
            "key_code": "volume_decrement"
          },
          "to": [
            {
              "key_code": "volume_decrement",
              "modifiers": [
                "left_shift",
                "left_option"
              ]
            }
          ]
        },
        {
          "type": "basic",
          "from": {
            "key_code": "volume_increment"
          },
          "to": [
            {
              "key_code": "volume_increment",
              "modifiers": [
                "left_shift",
                "left_option"
              ]
            }
          ]
        },
        {
          "type": "basic",
          "from": {
            "key_code": "f11",
            "modifiers": {
              "mandatory": [
                "fn"
              ]
            }
          },
          "to": [
            {
              "key_code": "volume_decrement",
              "modifiers": [
                "left_shift",
                "left_option"
              ]
            }
          ]
        },
        {
          "type": "basic",
          "from": {
            "key_code": "f12",
            "modifiers": {
              "mandatory": [
                "fn"
              ]
            }
          },
          "to": [
            {
              "key_code": "volume_increment",
              "modifiers": [
                "left_shift",
                "left_option"
              ]
            }
          ]
        }
      ]
    }
  ]
}
```
You can do that by first copying the JSON from above, then running the following command in you terminal:
```bash
pbpaste > ~/.config/karabiner/assets/complex_modifications/fine_volume_control.json
```
It should then show up as an option in your Karabiner settings under Complex Modifications -> Add rule.
Hope it helps! 🙂

mrelwood:
It didn't work directly, but I was able to slightly modify that to make it work! Finally! Thank you so much!!!

[...]

PR #154, opened 2017-10-14
Fine volume controls by default

bingocaller:
As requested in #153.  
Remap regular volume controls (either with or without `fn`) to fine volume controls (i.e. `Option` + `Shift` + Volume Up/Down).

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -24,7 +24,7 @@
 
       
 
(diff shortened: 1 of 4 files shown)


apm1467:
Thanks!

In macOS, the key names `f11` and `f12` are recognized as `volume_decrement` and `volume_increment` by the system. The `fn + f11` and `fn + f12` are the real `f11` and `f12`.

Normally, press the volume keys on Mac keyboard will send `f11` and `f12`; press Fn + the volume keys will send `fn + f11` and `fn + f12`.

This behaviour can however been changed in macOS System Preferences -> Keyboard. In that case, press the volume keys on Mac keyboard will send `fn + f11` and `fn + f12`; press Fn + the volume keys will send `f11` and `f12`.

So, if we want to send real `f11` and `f12` keys, we should actually send `fn + f11` and `fn + f12` in Karabiner-Elements. That will be recognized by the system as  real `f11` and `f12`, not volume keys.

I will change this rule a little bit, so the user can still send real `f11` and `f12` keys.

bingocaller:
@apm1467 
Thanks! 😄

PR #158, opened 2017-10-17
add a complex mod to switch the square brackets with the parens,

henryirvine:
putting the frequently used paren (at least for us lisp programmers) on a directly accessible key.


tekezo:
Thanks!

PR #161, opened 2017-10-18
Added swap_yen_and_backslash rule.

cayal:
Instead of using the US Keyboard layout, I use Japanese input with Romaji (that way, I can just ctrl+Space and get ひらがな／カタカナ。The Romaji layout hides the backslash under "Alt+\", though, and the lone "\" yields ¥. This config swaps the two.

Eventually it would be cool to get the Romaji layout to match the US keyboard layout exactly (I think there are a few other differences with special characters), but "¥"/"\" is the only difference I've actually noticed.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -24,7 +24,7 @@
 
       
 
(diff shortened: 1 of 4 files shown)


apm1467:
Thanks!

PR #163, opened 2017-10-20
add src/json/command_esc_to_command_grave_accent.json

BOZ88:
when use HHKB, grave_accent_and_title button is too far to use.


apm1467:
Thanks!

PR #166, opened 2017-10-20
A few Modern Space Cadet tricks

bingocaller:
A few modifications based on [Steve Losh's A Modern Space Cadet article](http://stevelosh.com/blog/2012/10/a-modern-space-cadet/), not including the Greek and Math layers.  
It also includes a Caps Lock to Escape/Control remapping, even though that is already present as a modifier in the "Change caps_lock key" group.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -22,9 +22,9 @@
         <a href="https://github.com/pqrs-org/KE-complex_modifications">GitHub</a>
       </div>
 
+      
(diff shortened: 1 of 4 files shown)


apm1467:
Thanks!

bingocaller:
No problem! I'm using the bindings myself, so I thought I might as well share 'em! 😁

PR #170, opened 2017-10-26
Add Korean PC Keyboard rules

RianGray:
I hope I am following the directions correctly to add a complex modifications rule. 

I am currently using a Korean PC keyboard, and the layout is slightly different from English QWERTYs. Korean keyboards come with two keys that change language input and change korean into hanja. In KE, they are each identified as lang1 and lang2, but do not work as intended. With this rules, each keys will work on Mac as well, as they will be modified to input corrensponding Mac key combinations instead.


AlexeyPopov:
1. Add `file_import_panel` to `src/index.html.erb`.
2. Run `make` in terminal.

RianGray:
Sorry about the mess. I'm still not used to Github. I just finished two things you mentioned.

AlexeyPopov:
LGTM
Thanks!

PR #172, opened 2017-10-30
Updates outdated modkeys (right alt) to (right option) for dvorak layout

michaelfresco:
This PR tries to fix the issue I had where the `option` key does no longer work properly with the Dvorak layout on Karabiner Elements. I changed the the modifier keys from `right_alt` to `right_option`.

@Reviewers: I suspect that the file `src/json/dvorak_layout.json.erb` is used to automatically generate the `dvorak_layout.json` so I made changes there as well.


apm1467:
That's totally right. Thanks!

PR #175, opened 2017-11-01
make change_grave_accent_into_escape work in more apps

adamfeuer:
This makes the change_grave_accent_into_escape work in more apps (like PyCharm, WebStorm, Chrome, etc.) by making it use `to` rather than `to_if_alone`.

I don't know why this works better, but it does.

### DIFF
--- a/docs/json/change_grave_accent_to_escape.json
+++ b/docs/json/change_grave_accent_to_escape.json
@@ -7,6 +7,17 @@
         {
           "type": "basic",
           "from": {
+            "key_code": "grave_accent_and_tilde"
+          },
+          "to": [
+            {
+              "key_code": "escape"
+            }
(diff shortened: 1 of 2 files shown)


tekezo:
You should use Simple Modifications for this purpose.
https://pqrs.org/osx/karabiner/document.html#configuration-simple-modifications

adamfeuer:
@tekezo A simple modification works if I only want to change one key combination, but this file does several things - changes `grave_accent` into `escape` if it is pressed alone, but makes `command-grave_accent` work as normal, and `shift-grave_accent` emit `~` also as normal. 

I didn't see a way to do that with simple modifications. I tried a combination of simple and complex for this, but that didn't work.

Thank you for accepting and merging this!

PR #176, opened 2017-11-01
Modern Space Cadet rev2: more granular settings

bingocaller:
I split the 'Better Shifting' rules into three separate rulesets because I realised users might not want all three settings at the same time.


tekezo:
Thanks!

PR #189, opened 2017-11-07
Master

Drazcmd:
Adding some personal rules I’ve been playing around with the last month.


Drazcmd:
(probably will want to squash if/when merging this in)

tekezo:
Thanks!

PR #204, opened 2017-11-13
fix command / control bug in VMs (Rule virtual_machine.json)

filipre:
Hi, 

there was an issue in the `virtual_machine.json` rule, see issue https://github.com/tekezo/Karabiner-Elements/issues/1091. The configuration I propose works well for me in my virtual machine using Ubuntu.
I think the problem was that `command` and `control` do not exist as a type but only `left_control`, `right_control`, `left_command`, and `right_command`. 

Cheers
René

### DIFF
--- a/docs/json/virtual_machine.json
+++ b/docs/json/virtual_machine.json
@@ -7,7 +7,7 @@
         {
           "type": "basic",
           "from": {
-            "key_code": "control",
+            "key_code": "left_control",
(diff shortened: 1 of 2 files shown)


apm1467:
Thanks!

It's kind of weird. When using as modifier keys, `command` and `control` do work. It seems like when directly using as `"key_code"`, they are not recognized by KE.

PR #210, opened 2017-11-18
Add rule to map eisuu or kana combined with other keys to option

ypresto:
I often map eisuu and kana keys to option keys, since long ago for vim (make option+any key to ESC+that key).

I have used Hammerspoon for this, but is not complete because it always behaves like (Karabiner-Elements') "lazy" option enabled.
Not making this "lazy" is important because holding option key is mapped to force quit in dock, some hidden functions in menu or etc...

Thanks for creating Karabiner and Karabiner-Elements, they totally saved my life time..!

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1382,11 +1382,11 @@
         <div class="panel-body">
               <div class="panel panel-default">
       <div class="panel-heading">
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

PR #239, opened 2017-12-02
Add New Poker II (Backtick, tilde & i/j/k/l as arrows)

ethan605:
There're some changes on iKBC New Poker II (2017), compared to Pok3r, that may affect developer's experience on MacOS:

  * Fn + Esc to insert backtick (`), now can use Capslock + Esc
  * Fn + Shift + Esc to insert tilde (~), now can use L_Shift + Esc
  * Fn + WASD as arrow keys, now can use L_Control + IJKL


apm1467:
Thanks!

PR #251, opened 2017-12-10
Add Half-QWERTZ one-handed layout

rhuitl:
This implements a QWERTZ layout that is usable with one hand only by holding the space bar. Inspired by https://github.com/garrow/karabiner-half-hand-qwerty.

As far as I could see, there is no equivalent for ModifierFlag::EXTRA1 in Karabiner Elements, so I used the right control key instead.

This is for a German keyboard layout, adapting it for QWERTY should be straightforward.

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -170,6 +170,9 @@
         },
         {
           "path": "json/neo2.json"
+        },
+        {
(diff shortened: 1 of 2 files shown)


apm1467:
Thanks!

PR #254, opened 2017-12-13
feature control_fn_to_media_control

bbvch13531:
In a 68-keyboard, there are serveral discomfort using media control.
Add feature control+fn to media control.


apm1467:
Thanks!

PR #256, opened 2017-12-16
disable power key

benng127:
I want to disable power key of macbook air due to dysfunction of power key.

### DIFF
--- a/docs/json/disable power key
+++ b/docs/json/disable power key
@@ -1,0 +1,11 @@
+{
+    "type": "basic",
+    "from": {
+        "key_code": “power”,
+        "modifiers": {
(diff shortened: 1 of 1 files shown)


apm1467:
You can use Simple Modifications to do that.

benng127:
I tried, "power" (from PC keyboard) to disable (from virtual keyboard). Didn't work. My power button/keyboard has hardware issue and would constantly shut down my Mac and then turn on. Unplug keyboard worked but I only need to disable power button.

tekezo:
Unfortunately, Karabiner-Elements cannot interrupt the shutdown signal.
(I think any software should not interrupt it.)

PR #267, opened 2018-01-04
Vi mode with caps lock

sabberworm:
Similar to vi mode but uses caps lock as a modifier.

Also allows scrolling via caps_lock + shift.


tekezo:
Thanks!

PR #268, opened 2018-01-04
Personal rules (@kbdluv)

kbdluv:
First commit.

- Post Escape if Command_L is pressed alone.
- Post Escape if Command_R is pressed alone.
- Post left_control if Spacebar is held down.
- Post grave_accent_and_tilde if Caps Lock is pressedis pressed alone, left_shift if held down.
- Post return_or_enter if Enter is pressed alone, right_shift if held down.

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -324,6 +324,9 @@
         {
           "path": "json/personal_yqrashawn.json",
           "extra_description_path": "extra_descriptions/personal_yqrashawn.json.html"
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

kbdluv:
Thank _you_ for this awesome product! And for the commit access too 😊. Should I always rebase and push to master when editing my personal rules instead of doing a PR?

PR #275, opened 2018-01-11
Disable Cmd+F1 Toggle Screen Mirroring Shortcut 

ptim:
(map to display_brightness_decrement)

See also https://discussions.apple.com/thread/2793101


tekezo:
Thanks!

PR #281, opened 2018-01-20
Adds rule "Equals + Delete to Forward Delete"

toblich:
This adds a rule that maps `=`+`Del` to `Forward Del`. It is equivalent to a rule that existed in the old Karabiner (before Karabiner-Elements, in pre-Sierra OS).

> This is my first PR to any Karabiner-related repos, I've read the README.md and tried to follow the style of other files but please let me know if I've violated any conventions 🙂


apm1467:
Thanks!

PR #289, opened 2018-02-04
Improvements of the Neo2 configurations

memo33:
Mod4 is now implemented using variables instead of overloading
right_command, which allows for using cmd + mod4 shortcuts, and also
allows for more involved private modifications that need the
right_command flag to be unset.

Fixes capslock incorrectly being triggered in layer 4.

Fixes keypad numeric operators, see jgosmann/KE-complex_modifications#2.


tekezo:
Thanks!

PR #298, opened 2018-02-20
app-specific convert hangouts mic/video mute to zoom.us

logikal:
I've been using Hangouts for over 6 years now, but my org is moving to Zoom. there's a lot of muscle memory built up, and this helps with that.

While Zoom.us is the frontmost app:
convert cmd-d to cmd-shift-a
convert cmd-e to cmd-shift-v

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -125,6 +125,9 @@
       "id": "application-specific",
       "files": [
         {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #302, opened 2018-02-22
A German Programming Mapping

wpp:
Common programming characters like {} and [] are mapped horribly
on a German keyboard. Hope this helps some folks.

I think this used to be available in the old Karabiner.


wpp:
Yup found an old screenshot, this mapping is basically replicating the bottom section: 
karabiner

tekezo:
Thanks!

PR #304, opened 2018-02-23
Travis improvements

logikal:
Try adding a linux build to travis-ci.

If we can build on linux VMs, we may be able to get our tests run much faster.


tekezo:
Thanks!

I guess you want to switch unit testing environment to linux.
(I understand matrix does not help to reduce time in this case. Is it misunderstand?)

But, I think that running unit testing on macOS with system rvm is reasonable to guarantee the build works fine on macOS.
(This project users are almost using macOS since Karabiner-Elements runs only on macOS.)
Thus, we should keep the osx environment.

logikal:
Agreed. I'd run into an hour long wait for an osx executor and rage-built this PR. Will close.

Thanks!

PR #305, opened 2018-02-26
Caps lock to left ctrl when pressed with other keys

cflann:
For those of us who actually want to use caps-lock sometimes.


tekezo:
Thanks!

PR #308, opened 2018-03-02
Update "Equals+Del" rule to use new "simultaneous" feature

toblich:
The older version of this modification, which I added in #281, had some issues when modifiers were pressed. This new version uses the new "simultaneous key presses" to work as expected. 

The code is actually the one in the [documentation](https://pqrs.org/osx/karabiner/json.html#typical-complex_modifications-examples-simultaneous-equal-delete-to-forward-delete), but I didn't see it in the list of complex modification that could be added directly from the "Preferences", therefore I'm updating my old code to use this new one so anyone can import it in an easier manner (let me know if I missed some place).

### DIFF
--- a/docs/json/equals_del_to_forward_del.json
+++ b/docs/json/equals_del_to_forward_del.json
@@ -2,50 +2,28 @@
   "title": "Equals + Delete to Forward Delete",
   "rules": [
     {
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

PR #309, opened 2018-03-05
erb2json: add kitty to terminal_bundle_identifiers

seanfarley:
As the subject line says, this is just adding [kitty](https://github.com/kovidgoyal/kitty) to the list of known terminals.


tekezo:
Thanks!

PR #312, opened 2018-03-08
Add "Ctrl+[を押したときに、escキーと英数キーを送信する" (Send Escape and Eisuu when Ctrl+…

shinkbr:
This PR will extend the vim-like `Ctrl+[` -> `Escape AND Eisuu` behavior to outside of vim, say when browsing the web or using other text editors.
AFAIK there is currently no way to achieve this with existing modification rules.

### DIFF
--- a/docs/json/japanese.json
+++ b/docs/json/japanese.json
@@ -183,6 +183,66 @@
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #313, opened 2018-03-09
remove Launcher Mode v3, Mouse Keys Mode v3 from personal_tekezo.json

tekezo:


threescales:
Hello,Why close it? I like this function very much......

threescales:
This is great!!!
But why can I just move my mouse by press hjkl? Dosen't need 'd'.

threescales:
I generate a package and the problem is solved.It's wonderful！

PR #315, opened 2018-03-12
Ctrl+6 to Ctrl+Shift+6 for vim buffer switching

sergii-bond:
I haven't found the existing code, so wrote my own based on other examples. The key mapping is intended to be used in vim to switch between the previous/next buffers. Ctrl+6 doesn't not work in MacOS directly. Luckily vim supports Ctrl+Shift+6 combination to have the same function as Ctrl+6.

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -60,6 +60,9 @@
           "path": "json/vim_style_escape.json"
         },
         {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #324, opened 2018-03-28
Add Gamer Cursor keys [Fn+WASD]

PJWalker:
Built to cater to habits from my KBParadise V60, but may be of use to others as well.


tekezo:
Thanks!

PR #327, opened 2018-03-30
Mouse button 4 (side button on Microsoft mice) to double-click

GeoffreyBooth:
I feel like there should maybe be a shortcut for double-click.


tekezo:
Thanks!

PR #330, opened 2018-04-12
Add emacs bindings to kill word and line

abstiles:
This addresses an issue previously filed in tekezo/Karabiner-Elements#1215 (missing Ctrl+w shortcut) as well as an additional shortcut (Ctrl+u) I also used to use with the old Karabiner.

### DIFF
--- a/docs/json/emacs_key_bindings.json
+++ b/docs/json/emacs_key_bindings.json
@@ -929,6 +929,148 @@
               ]
             }
           ]
+        },
+        {
+          "type": "basic",
+          "from": {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

I moved them into "Bash style Emacs key bindings since they are not Emacs native.

https://github.com/pqrs-org/KE-complex_modifications/commit/987be0f6dc037a32d6e56758610ece91ca9c93d8#diff-b7fc5af635f8e809b7e7a74d0e66168c

PR #331, opened 2018-04-15
r_cmd and r_shift to cmd + tab

BennyHarv3:
Adds a complex modification for mapping right_cmd and right_shift to command and tab. Base modifier is right_shift, base to modifier is tab.


tekezo:
Thanks!

PR #332, opened 2018-04-18
Numeric Keypad Mode: Add option to use the left command as spacebar

vinceau:
The use case is when you want to type many space separated numbers like phone numbers without lifting and re-pressing the tab key.


tekezo:
Thanks!

PR #339, opened 2018-04-19
Numeric Keypad Mode: Add alternative layout.

vinceau:
This adds an option to use right_command as zero in order to preserve the spacebar for actual spaces.

I'm not sure if this is the optimal layout for having a zero key while preserving the spacebar. Open to suggestions.


tekezo:
Thanks!

tekezo:
I think your layout is reasonable while we are using uio/jkl/m,. as keypad numbers.

PR #341, opened 2018-04-29
Printscreen to clipboard

GeoffreyBooth:
Allow mapping <kbd>PrintScreen</kbd> to either take a screenshot and have that saved to a file on the desktop, or have it copied to the clipboard.


tekezo:
Thanks!

PR #344, opened 2018-04-29
Command_R + Return to backslash

gracjan-grala:
#### `Command_R` (⌘) + `Return` (⏎) to `Backslash` (\\) and `Command_R` (⌘) + `Shift` (⇧) + `Return` (⏎) to Pipe (|)

This rule is handy when you are used to or would like to use [the US English Layout](https://support.apple.com/library/content/dam/edam/applecare/images/en_US/keyboards/english_notebook.png) but have one of [the layouts with the big Return key](https://support.apple.com/library/content/dam/edam/applecare/images/en_US/keyboards/en_international_notebook.png).

Then the backslash key next to the Return becomes really annoying.

You can remap the `backslash (\)` key to `return_or_enter` with a _Simple Modification_ and then enable this _Complex Modification_ to make your backslash and pipe keys accessible when you press your right Command.

### DIFF
--- a/docs/extra_descriptions/cmd_return.json.html
+++ b/docs/extra_descriptions/cmd_return.json.html
@@ -1,0 +1,32 @@
+<style>
+.monospace {
+  font-family: monospace;
+}
+</style>
+
+<p>
+  This rule is handy when you are used to or would like to use
+  <a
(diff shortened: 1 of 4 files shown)


tekezo:
Thanks!

PR #346, opened 2018-04-30
Numeric Keypad Mode: Add Caps Lock and Escape as trigger keys

vinceau:
Add option to use different keys as trigger keys.

## Use case
When entering data into a spreadsheet, the Tab key is often used to navigate to the next cell horizontally. Currently if you wanted to enter numbers in several cells, you would have to press and release Tab over and over (as the trigger, and then as Tab for going to the next cell). If a different key is used as a trigger you don't suffer from this issue.

Caps Lock is a good key for a trigger however it is often bound to something else (such as escape for Vim users). So add both Caps Lock and Escape as options for trigger keys.

## Concerns
This current implementation requires users to enable two separate modifications, one for selecting the trigger key, and another for selecting the number pad layout (either right_command or spacebar as the key for 0). I've tried making this clear in the extra_descriptions but I don't know if this is too complicated for users who would rather only need to enable one thing as was the case previously.

### DIFF
--- a/docs/extra_descriptions/numeric_keypad.json.html
+++ b/docs/extra_descriptions/numeric_keypad.json.html
@@ -41,9 +41,10 @@
 </table>
 
 <ul>
-  <li>Hold down the trigger key (Tab) then press one of the keys to use a numeric key</li>
-  <li>Press and releasing tab acts as a normal tab key</li>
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #351, opened 2018-05-02
Add backslash-yen swapping for JIS keyboards

tanakahisateru:
I can't modify JIS keyboard's **¥** key using Karabiner-Element with existing rule set "Swap ¥ and \ for Japanese Romaji input on US Keyboard".

Please add the rule for also JIS keyboards. Differ from US keyboards, **¥** is mapped to the virtual key code "international3" for JIS keyboard.

See details why/how we do:
- https://qiita.com/tanakahisateru/items/d8a67ad89c07baf8c0da
- https://qiita.com/kunit/items/f6ec7e159d93002bfc48

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -244,6 +244,9 @@
           "path": "json/swap_yen_and_backslash.json"
         },
         {
+          "path": "json/swap_yen_and_backslash_jis.json"
+        },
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

PR #362, opened 2018-05-15
add control and left_option modifiers to gamer_keys

asny23:
added rules to "Gaming cursor keys"

- control + w/a/s/d to arrow keys
- left-option + w/a/s/d to arrow keys


tekezo:
Thanks!

PR #363, opened 2018-05-17
Screen saver hotkeys (that can lock screen)

sshymko:
Two options to launch screen saver:
1. Option + L
2. Control + Command + Delete

Note: macOS can be further configured to immediately lock the screen:
System Preferences > Security & Privacy > Require password: immediately
os-x-require-password-immediately


tekezo:
Thanks!

PR #364, opened 2018-05-17
Lock screen hotkeys on macOS High Sierra 10.13+

sshymko:
Two options to lock screen on macOS High Sierra 10.13+:
1. Option + L
2. Control + Command + Delete

Related to #363


tekezo:
Thanks!

Issue #361, opened 2018-05-14
Karabiner Complex Modification Of "Programmer Dvorak-CmdQwerty Keyboard" has a bug that I can't locate (and therefore can't fix). [Mostly Resolved]


leonzalion:
Currently, I use this keyboard layout every day and I recently discovered a bug that prevented me from typing properly when typing some code. When you press the equal sign while the plus sign is pressed (on QWERTY equal sign is 6 and plus sign is 9), the equals sign won't be typed out, which prevents me from typing code that requires `+=` fast, for I need to lift up the plus sign before pressing the equal sign (but other key combinations like `!=` work perfectly fine). Looking through the code, I was unable to locate the origin of this bug, so I was wondering if anybody could help me with this and help me find the origin and maybe offer a solution to it, if it isn't just me who's experiencing this (Also, it's the same thing the other way around with `=+`, and another weird behaviour these two keys have is when you try typing `+++++=====` using key repeat (holding plus so that it repeats and then – without lifting the plus sign, use another finger to hold the equals sign), you'll notice that when it switches between the plus signs and the equal signs it doesn't stop like it would with some other key combination like `+++++)))))` – try it).

Link to code: https://github.com/pqrs-org/KE-complex_modifications/blob/master/docs/json/programmer_dvorak_cmd_qwerty_layout.json

P.S. There is another bug with the layout that is easy to fix, but just to bring it to light – when you press the dollar sign (grave_accent_and_tilde on QWERTY) with shift pressed, it becomes a grave accent instead of a tilde like it should be.

leonzalion:
Just found out the cause of this, and it's not just the layout's fault. When you do the action that I gave with two keys that share the same button on a QWERTY layout (for example, plus and equals – you get plus with shift and equals without shift), this unwanted result will occur. The reason is that karabiner treats it as simply lifting up shift when you switch between the two keys, therefore not repeating the key again, but merely lifting up the shift so that the wanted key is 'pressed,' but doesn't repeat and is therefore not displayed. To see this for yourself, some other examples include tilde and grave accent `\`~`, and `]}` (they share the same button on a QWERTY layout). Therefore, this is not a flaw of the complex modification code, but rather the software itself (You could test this with Karabiner Event Viewer, where doing one of these actions simply results in the shift being lifted up, but the original key never being pressed twice).

leonzalion:
[...]

PR #366, opened 2018-05-20
fix “Programmer Dvorak-CmdQwerty Keyboard”

apm1467:
fix issue #361


bradywatkins-ooh-n:
Hi @apm1467 , I was wondering if you could release the "Programmer Dvorak" keyboard layout *without* CmdQwerty. Thanks

PR #370, opened 2018-05-25
added rotate_left_modifiers

louispan:
This rotates left ctrl to command, command to option, option to control; resulting in physical key mapping similar to Windows/Linux; supporting application switching with option+tab.


apm1467:
Thanks!

PR #374, opened 2018-05-31
Adding Sleep Monitor Rule for Keyboards without Eject 

RianGray:
I've created a modification rule for keyboards without an eject button. Currently Mac's sleep monitor shortcut requires eject button on the keyboard, so this rule adds another way to trigger the shortcut without an eject key.

Let me know if I missed anything. Thanks in advance!


tekezo:
Thanks!

PR #375, opened 2018-05-31
Change Won (₩) to grave accent (`) in Korean layout

e9t:
This PR remaps the dead key for the Won symbol to the grave accent in the Korean layout.
Should be useful for Korean developers.

(Work done with @shurain)


tekezo:
Thanks!

PR #377, opened 2018-06-03
switch to standard function keys in virtual env

### LINKED ISSUE #376: [Request] switch to standard function keys in virtual env
I want to standard function keys in virtual environment, i.e. use normal f1, f2 without fn. However, according to [Input event modification chaining], modification for function keys in `Complex Modifications` will be changed in `Function Keys Modifications` anyway.

    Karabiner-Elements treats input events by the following order:

        1. Catch events from hardware.
        2. Apply Simple Modifications.
        3. Apply Complex Modifications.
        4. Apply Function Keys Modifications (change f1...f12 keys to media controls).
        5. Post events to applications via a virtual keyboard.

So, is there a parameter in `complex_modifications` to skip stage `Function Keys Modifications`? Or how can we implement this request?

[Input event modification chaining]: https://pqrs.org/osx/karabiner/document.html#event-modification-chaining

holymonson:
Thanks @tekezo . Close #376 .

### DIFF
--- a/docs/json/virtual_machine.json
+++ b/docs/json/virtual_machine.json
@@ -235,6 +235,1072 @@
           ]
         }
       ]
+    },
+    {
+      "description": "Use standard function keys in virtual machine/remote desktop",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "f1"
+          },
+          "to": [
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #378, opened 2018-06-04
Seperate left and right rules to swap command and control in virtual machine

holymonson:
No `right control` on macbook's keyboard, and `right command` is usually set as the host key. Swapping the left only is good enough.


tekezo:
Thanks!

PR #382, opened 2018-06-05
Make pc-style Home/End work in git-gui and Jetbrains IDEs

arichardson:
In git-gui home/end is handled and cmd+arrow only moves by 1 character.
Jetbrains IDEs contain an embedded terminal which doesn't like cmd+arrow.


tekezo:
Thanks!

PR #391, opened 2018-06-23
Add half-qwerty emulation

void:
Add new alternative keyboard layout that emulates the behavior of the [half-qwerty keyboard](http://half-qwerty.com/)


apm1467:
Thanks!

PR #404, opened 2018-07-12
Add rule: exchange single and double quote

ybiquitous:
This adds a rule of exchange quote (single and double), which is very useful for US keyboard layout.

- `'` (single quote) and `"` (double quote) is same key in US layout (such as semicolon and colon).
- For people who type often "double" than "single" in programming, I think this rule is very useful.

image

Thanks.


apm1467:
Thanks!

PR #408, opened 2018-07-16
add supplemental rules for diamond cursor with right command.

ntaoo:
I would like to merge my private configs into public.

https://github.com/ntaoo/right_hand_diamond_cursor_plus

These are intended to be supplemental for Diamond Cursor right_command rules.

* Change right_command+u/m/o/. to page up/page down/home/end
* Change right_command+;/h to delete/forward delete


apm1467:
Thanks!

PR #410, opened 2018-07-19
Vim pedal for modes

max-sixty:
This allows you to depress a pedal to enter edit mode, and then release it to return to command mode


apm1467:
Thanks!

PR #413, opened 2018-08-07
Add modifications for Visual Studio Code

bsiegel:
- Adds proper back/forward navigation using mouse side buttons.
- Adds "Go to file" on double tap of shift key (like IntelliJ/Android Studio) - requires adding a keybinding in VSCode, see the extra descriptions.


apm1467:
Thanks!

PR #416, opened 2018-08-15
Create new template to remap escape key on Mac w. TouchBar

stanislasdrg:
Remap MacOs escape key to left_shift and left_shift + escape to escape

### DIFF
--- a/docs/json/mac_touchbar_escape
+++ b/docs/json/mac_touchbar_escape
@@ -1,0 +1,32 @@
+{
+  "title": "MacOs Touchbar Escape Key",
+  "rules": [
+    {
+      "description": "Mac Touchbar Escape key (escape as left_shift and left_shift + escape as escape)",
+      "manipulators": [
+        {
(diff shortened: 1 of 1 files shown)


apm1467:
What's the intention of this rule? If escape is mapped to left_shift, there will be no escape key on the keyboard. How would you press the key combination shift + escape then?

stanislasdrg:
The intention was to be able to map the sole escape key to a useless key, while keeping the ability to escape. It is a bit hacky, but it does work on my end. Here is [a short demo video](https://streamable.com/91bhy) I made.

stanislasdrg:
Thanks, yes that's exactly what I am trying to do. The Touchbar's design, favorises typing on escape by accident. I was trying to fix that behaviour and was surprised no KE rule existed, despite KE being the recommended solution.

apm1467:
This should work fine: https://pqrs.org/osx/karabiner/complex_modifications/#escape

Issue #427, opened 2018-08-31
Citrix not supported in the Virtual Machine complex_modifications rules


therk:
I am using Citrix to access windows and need the Ctrl -> Command remapping. 
Per virtual_machine.json, it is not supported. 
Need to add "^com\\.citrix\\.XenAppViewer$"

PR #428, opened 2018-08-31
#427 Added Citrix support to the Virtual Machine complex_modifications

therk:
Added Citrix support into Virtual machine by adding "^com\.citrix\.XenAppViewer$"


apm1467:
Thanks!

PR #430, opened 2018-09-06
Add Programmer Dvorak CmdQwerty toggle JSON.

senbrow:
Unlike the _layout.json variant, which also includes a full mapping from
QWERTY->Programmer Dvorak, this ruleset sits on top of the "native" Programmer
Dvorak layout installed at the operating system level and dynamically
toggles to QWERTY when either command key is pressed.

This has the advantage of letting the user use the standard operating
system layout controls while also preserving QWERTY command shortcuts.

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -224,6 +224,10 @@
           "path": "json/programmer_dvorak_cmd_qwerty_layout.json"
         },
         {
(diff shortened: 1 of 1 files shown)


senbrow:
> So the user needs to first install this package before using your rule, right?

Exactly. Once the driver package is installed, these rules will simply toggle the OS-wide layout to it when the command key is pressed and toggle back when it is released.

PR #432, opened 2018-09-08
Add rule: use option+1/2/3/4 to switch languages (English, Chinese, Japanese)

hanezu:
This adds a rule for switching languages by

- option + 1 for English
- option + 2 for Chinese
- option + 3 for Japanese, Hirakana
- option + 4 for Japanese, Katakana

It is very useful for English + Chinese + Japanese users (switching in a role is inconvenient especially for people constantly using more than two languages.
It should work fine for any Chinese/Japanese input method.
Thanks.

PS: excuse me if I put it in the wrong place but I'm not sure whether it should be Personal Settings or Miscellaneous.

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -491,6 +491,9 @@
         },
         {
           "path": "json/save_pinky_fingers.json"
+        },
+        {
+          "path": "json/switch_en_cn_ja_katakana.json"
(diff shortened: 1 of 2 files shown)


apm1467:
Thanks!

PR #435, opened 2018-09-20
Add modifier keys and VNC applications to virtual_machine

ehlers:
Two modifications to virtual_machine:

- Allow modifier keys for function keys, so Alt-F4 works as expected.
- Add VNC to the list of applications. The list of bundle identifiers comes from appdef.xml of the old Karabiner.


apm1467:
Thanks!

PR #438, opened 2018-09-24
Add rule: Change fn + grave_accent_and_tilde to esc

youknowone:
This rule doesn't change default behavior of grave_accent_and_tilde but adds hardware esc variant with fn key.


apm1467:
Thanks!

PR #441, opened 2018-09-30
Add soft-meta for Alacritty

eraserhd:
Alacritty has an awkward key-mapping facility which does not correctly
work around diacritics, so it is not completey possible to map, say,
Option-i to Meta-i.  This sends Escape+i, which is the terminal code
for Meta-i, making Option a workable meta key.


apm1467:
Thanks!

lambdalisue:
Thanks for sharing that 💯

PR #447, opened 2018-10-18
add "OptionTab" into doc.

shell-autonomic-sh:
Use Option + number to Switch Tabs
working in chrome and firefox.


apm1467:
Thanks!

PR #449, opened 2018-10-24
Added rule to allow activation of the standard fn keys

Jendker:
This allows activation of standard function keys with fn button
on per profile basis. For now it is not possible, because the
activation of F1, F2... buttons is a global settings. So settings
the F1, F2 and so on buttons to F1, F2... and using this rule the
inversion of the behaviour is possible on per profile basis.

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -352,6 +352,9 @@
         },
         {
           "path": "json/command_r_and_shift_r_to_command_tab.json"
+        },
+        {
+          "path": "json/standard_function_keys_on_fn.json"
         }
       ]
     },
--- a/docs/json/standard_function_keys_on_fn.json
+++ b/docs/json/standard_function_keys_on_fn.json
@@ -1,0 +1,197 @@
+{
+  "title": "Standard function keys using the fn key",
+  "rules": [
+    {
+      "description" : "Standard function keys using fn key - allows adding this feature just for specific profiles.",
+      "manipulators" : [
+        {
+          "type" : "basic",
+          "to" : [
+            {
+              "key_code" : "display_brightness_decrement"
+            }
+          ],
+          "from" : {
+            "key_code" : "f1",
+            "modifiers" : {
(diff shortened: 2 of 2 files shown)


apm1467:
Thanks!

ajdajd:
Hello! I have trouble making this Complex Modification to work. Does adding this to a profile invert the global rule?

Jendker:
I added the rules to make it work like this (inversion of the global rule for profile), it is right now a pull request: https://github.com/pqrs-org/KE-complex_modifications/pull/451

Here if you have Macbook Pro with Touch Bar It works better if you activate checkbox in Function Keys menu and select 3rd rule ("Keep normal media keys..."), because for the 2nd option and media keys as default I needed to add Fn modifier what makes the Touch Bar blink during key press. If you don't have Touch Bar there are no restriction, just select option 2 or 3 depending on the state of the checkbox in Function Keys menu (2nd rule for checkbox disabled, 3rd otherwise).

You could try to download the file and add it to ~/.config/karabiner/assets/complex_modifications to make it work right now.

Jendker:
To make it work right now you would need to assign F1 to F1, F2 to F2 etc. in Function Keys menu.

I added the rules to make it work like this (inversion of the global rule for profile), it is right now a pull request: https://github.com/pqrs-org/KE-complex_modifications/pull/451

Here if you have Macbook Pro with Touch Bar It works better if you activate checkbox in Function Keys menu and select 3rd rule ("Keep normal media keys..."), because for the 2nd option and media keys as default I needed to add Fn modifier what makes the Touch Bar blink during key press. If you don't have Touch Bar there are no restriction, just select option 2 or 3 depending on the state of the checkbox in Function Keys menu (2nd rule for checkbox disabled, 3rd otherwise).

You could try to download the file and add it to ~/.config/karabiner/assets/complex_modifications to make it work right now.

ajdajd:
Thanks! I missed this.

PR #453, opened 2018-11-05
Add additional Finder rules

### LINKED ISSUE #450: Remap delete key to go to previous folder / Fn+delete key to move to trash / F2 to rename files
Hi! I just found the rule that allows to open files/folders with return key, so I'd like to know if it's possible to add rules mentioned in the title to make it more convenient if you work simultaneously on mac and windows.

Thanks!

medusalix:
Adds the following rules to the Finder:

- Use F2 as Rename
- Use Fn+Delete as Move to Trash
- Use Backspace as Go to Previous Folder

Resolves #450

### DIFF
--- a/docs/json/finder.json
+++ b/docs/json/finder.json
@@ -79,6 +79,30 @@
       ]
     },
     {
+      "description": "Use F2 as Rename",
+      "manipulators": [
+        {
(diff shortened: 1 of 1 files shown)


apm1467:
Thanks!

PR #457, opened 2018-11-27
Append Command_R send KANA and Control_L send EISUU rule

mongolyy:
I want to assign Japanese_EISUU to control L when exchange control L for command L


apm1467:
Thanks!

PR #459, opened 2018-12-01
add fn-hjkl mapping to arrow keys

bwasmith:


andmej:
Looks like `vi_style_arrows.json` is already doing this. Can you use that one instead?

andmej:
@apm1467 does this look good to you?

PR #460, opened 2018-12-05
Remap AZERTY "ù" key to "´" deadkey

nxadm:
Instead of using the azerty shortcut Alt+Shift+& for the ´ modifier, this scheme changes "ù" to the "´" deadkey (which is situated next to "´").


tekezo:
Thanks!

PR #464, opened 2018-12-28
Add rules for using Fn keys for copy & paste

aikinai:
F13 to Copy
F14 to Paste
⇧F14 to Paste and Match Style
F15 to Paste and Match Style

These are especially useful if you're on a full-sized Apple keyboard
(like the iMac Pro keyboard) or if you mouse on the left so your left
hand isn't always in position to hit the ⌘ shortcuts.

### DIFF
--- a/src/json/fn-copy-paste.json.erb
+++ b/src/json/fn-copy-paste.json.erb
@@ -1,0 +1,93 @@
+{
+  "title": "F13 to copy & F14 to paste",
+  "rules": [
+    {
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

I also updated groups.json.
https://github.com/pqrs-org/KE-complex_modifications/commit/32ba6e5fd5887b6f66a843ec7b1e892583c9cfe4

aikinai:
Great, thank you!

PR #468, opened 2019-01-02
Added a Window Resize Mode that can be used with 3rd party window resize apps

kamykaze:
Basically, when pressing the trigger key + one of the 10 mapped keys (which are mapped based on their position), you'll trigger a 'hardly' used shortcut key combination, which you can configure in another app (in my case, BetterTouchTool) to resize your current window.

Note: Karabiner Elements is unable to resize your mac windows by itself, so you do need another app to handle it. This simply makes it easier to trigger difficult shortcuts.


tekezo:
Thanks!

PR #470, opened 2019-01-25
Add right shift vi like

tom-power:
Hi there,

Another set of vi like bindings with right shift as modifier for those and some other keys.

I find it very useful and RSI repelling and wanted to share.

Thanks for the application, it rules \m/

Tom


tekezo:
Thanks!

PR #472, opened 2019-02-06
caps_lock to cmd+ctrl+opt+shift, caps_lock when pressed alone

Sesquipedalian:
This rule allows caps_lock to be used as cmd+ctrl+opt+shift when pressed in combination with any other key, but to still work as normal caps_lock when pressed alone.

Because this stays closer to standard keyboard behaviour than any other version of the "caps_lock to cmd+ctrl+opt+shift" rules, it is especially suitable for new users. Therefore, I have placed this rule first in the list in order to make it easier for new users to find it.

### DIFF
--- a/src/json/caps_lock.json.rb
+++ b/src/json/caps_lock.json.rb
@@ -31,7 +31,7 @@ def main
             ],
             'to_if_alone' => [
               {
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

PR #474, opened 2019-02-19
add Swedish Umlaut rule

LinusU:
Similar to the German Umlaut rule, but this one adds ä and ö (å already exists, which is alt + a)

Really helps as a Swede using the international English layout, as I can now type å, ä and ö without switching keyboard layout 😍


tekezo:
Thanks!

I updated groups.json.
https://github.com/pqrs-org/KE-complex_modifications/commit/c983f4b3c35d784f27b8a9c77961964e9997addb

LinusU:
Thanks! ❤️

PR #482, opened 2019-03-15
Rules to post CTRL from backtick, z, and slash

flyinggrizzly:
This PR adds config rules for alternative mappings for CTRL:

- holding ` (grave accent/tilde)
- holding `z`
- holding `/`

All of these keys when pressed solo emit their normal value.


tekezo:
Thanks!

PR #489, opened 2019-03-24
Added template for mapping FN + arrows to: page up/down, home, and end

patoui:
Added complex modifications template for:

FN + Up = Page Up

FN + Down = Page Down

FN + Left = Home

FN + Left = End


tekezo:
Thanks!

PR #490, opened 2019-03-27
Add workaround_broken_sw2tabesc.json

moberhuber:
This complex ruleset defines fn+d -> s , fn+e -> w
and so on, in order to work around my broken Macbook Air
keyboard which doesn't react on keys s,w,2,tab,esc,<
any more. By pressing fn+each desired key's right
neighbour, I can now reach all keys again which works
surprisingly well.

Restricted to operat on the Mac internal keyboard only,
and provided as an example in case somebody has a similar
problem.

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -567,6 +567,9 @@
         },
         {
           "path": "json/switch_en_cn_ja_katakana.json"
+        },
+        {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #492, opened 2019-05-07
Create VIMode.json

GenericMrBot:
hoping to get it to work only in chrome and Evernote.

### DIFF
--- a/docs/json/VIMode.json
+++ b/docs/json/VIMode.json
@@ -1,0 +1,1174 @@
+{
+  "title": "Vimium",
+  "rules": [
+    {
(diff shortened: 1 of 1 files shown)


tekezo:
You have to add `VIMode.json` to `groups.json`.

And please fix bundle_identifier like this.
https://github.com/pqrs-org/KE-complex_modifications/blob/5f1c741587584c44601ba39ed913264f65c86528/docs/json/pc_shortcuts.json#L182-L189

You can examine the bundle identifier on EventViewer.
https://pqrs.org/osx/karabiner/json.html#condition-definition-frontmost-application

PR #498, opened 2019-05-21
Add Microsoft Edge browser

alisonatwork:
Adding Microsoft Edge to the list of browsers. See: https://blogs.windows.com/msedgedev/2019/05/20/microsoft-edge-macos-canary-preview/

Currently this identifies as `com.microsoft.Edge.Canary` but presumably once it gets into standard insider preview and general release it will retain the `com.microsoft.Edge` prefix.


tekezo:
Thanks!

PR #499, opened 2019-05-21
More PC style shortcuts

alisonatwork:
There are a few shortcuts that I use fairly frequently in Windows that are missing from the PC style group. Here I add the following options:

* *Command-.* to pop up emoji picker
* *Command-S* to call up a Spotlight search
* *Command-Space* to switch inputs (e.g. from US Keyboard to Pinyin)
* *Control-Esc* opens Launchpad; this is an old Windows keypress to open Start Menu
* *Command-E* opens Finder

### DIFF
--- a/docs/json/pc_shortcuts.json
+++ b/docs/json/pc_shortcuts.json
@@ -1767,6 +1767,169 @@
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #471, opened 2019-01-29
Create F12 shortcut to open up Dev Tools Network

niij:
only for Mac/Firefox

PR #500, opened 2019-05-21
F12 developer tools proposals

alisonatwork:
Here is a proposal to modify the F12 developer tools keypress a little bit.

One important change I am proposing to the original version by @niij in #471 is to only have that rule take effect when Firefox is in the foreground. The reason is because it otherwise steals the F12 for other applications (notably Edge).

In ce9166f I am also suggesting we add a "generic" F12 mapping for all browsers. Chrome, Edge, Firefox and Safari all use Cmd-Opt-I to open the developer tools on the inspector tab, so this would work on all browsers as an alternative to the Firefox-specific option if you do not care about network tab.

Another option might be to combine these back together again so on Firefox it does Cmd-Opt-E and on Chrome/Safari it does Cmd-Opt-I. On Edge it doesn't matter because F12 works already.

Note: I did not `make rebuild` yet because this one might be different if #498 gets merged first.

### DIFF
--- a/src/json/f12-to-dev-tools-networking-mac-ff.json.erb
+++ b/src/json/f12-to-dev-tools-networking-mac-ff.json.erb
@@ -1,5 +1,5 @@
 {
-  "title": "Use F12 to open Dev Tools Network Tab (Mac + Firefox)",
+  "title": "Use F12 to open Dev Tools",
   "rules": [
     {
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

niij:
Why didn't you just create your own modification for opening Dev tools if that's what you needed?

As the filename/description stated, this is for quickly opening the network tab, not just Dev tools.

PR #505, opened 2019-06-20
Add rule: Figma Shortcut

kiepng:
### feature
- Post spacebar to Command+Shift+Ctrl if not press alone in Figma app.

This Shortcut operates the same  as the [Sketch Shortcut](https://github.com/pqrs-org/KE-complex_modifications/blob/master/docs/json/sketch-shortcut.json).

### DIFF
--- a/docs/json/figma-shortcut.json
+++ b/docs/json/figma-shortcut.json
@@ -1,0 +1,44 @@
+{
+  "title": "Figma Shortcut",
+  "rules": [
+    {
+      "description": "Post spacebar to Command+Shift+Ctrl if not press alone in Figma app.",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "spacebar",
+            "modifiers": {
+              "optional": [
+                "any"
+              ]
+            }
+          },
+          "to": [
+            {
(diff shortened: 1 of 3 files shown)


docs/json/figma-shortcut.json
```diff
@@ -0,0 +1,44 @@
...
+  "title": "Figma Shortcut",
+  "rules": [
+    {
+      "description": "Post spacebar to Command+Shift+Ctrl if not press alone in Figma app.",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "spacebar",
+            "modifiers": {
```
sigsevneo: You probably don't need this modifier block if you are already using the `to_if_alone` construct below?

tekezo: You can choose with/without `modifiers`.

- With `modifiers optional any`
  - The `spacebar` will always be changed with any modifiers such as command-space.
    The `spacebar` event will be fired after you release the physical space bar.
- Without `modifiers optional any`
  - The `spacebar` will be changed only no modifiers are pressed.
    When you type command-space, the `spacebar` event will be fired immediately.

PR #507, opened 2019-06-21
Add rules for AquaSKK workarounds.

qnighy:
Adds application-specific rules.

[AquaSKK](https://github.com/codefirst/aquaskk) is a Japanese IME and it's not fully compatible with JetBrains IDEs and Terminal/iTerm2. The semi-official workarounds using Karabiner are known (https://mzp.hatenablog.com/entry/2016/10/29/095435, https://mzp.hatenablog.com/entry/2015/03/15/213219) but they're in the old `private.xml` format. This PR is the port of the known workarounds.


tekezo:
Thanks!

PR #514, opened 2019-08-09
feat: Add workaround for auto-jumping or faulty keys

rakeshtembhurne:
These are personal rules. Disables faulty auto-jumping keys `left_option` and `9`. Adds new combo replacements for the characters `9` and `(`.


tekezo:
Thanks!

PR #516, opened 2019-08-18
Added missing Num Lock condition to keypad_7 key manipulation

imasaru:
The absence of the Num Lock condition restricts `keypad_7` to the `home` keycode regardless of whether Num Lock is enabled or disabled.


tekezo:
Thanks!

imasaru:
@tekezo Hey! 僕の提供を受けてありがとうございます。よろしくお願いします。

PR #517, opened 2019-08-21
Add default arguments to Karabiner.from_modifiers

mingaldrichgan:
`Karabiner.from_modifiers` is frequently called with `nil` and
`['any']` as the mandatory and optional modifiers, respectively.
Specifying a mandatory modifier and `['any']` as optional is also a
common use case. Thus, it seems reasonable to add `nil` and `['any']` as
default arguments for this method.

Running `make` produces no changes in the `docs/json` directory.


tekezo:
Thanks!

PR #526, opened 2019-09-01
Add helpful rules for the Matias Ergo Pro Keyboard + personal rules

mingaldrichgan:
The [Matias Ergo Pro](https://matias.ca/ergopro/) is one of the few
ergonomic mechanical keyboards on the market with an mostly-standard
Mac layout. Two quirks stand out:

1. The navigation keys (`home`, `end`, `page_up`, and `page_down`) are
where the `right_option` key is on a MacBook or Magic Keyboard.
2. The `right_control` key is where the B key *would* be *if* it were
on the right side of the split (which it isn't).

The following rules remap theys keys in a context-sensitive manner:

1. Change navigation keys to right_option if pressed with another key
2. Change right_control to b if pressed alone, or while typing

This commit also introduces some personal rules of mine, as well as the
`KeyCodes` and `Variable` utility classes used to build both rule sets
(and likely to be helpful for other future rules too).

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -506,6 +506,9 @@
       "id": "device-specific",
       "files": [
         {
+          "path": "json/matias_ergo_pro.json"
+        },
+        {
           "path": "json/typematrix_bepo_cut_copy_paste_fn_shortcuts.json"
         },
         {
@@ -566,6 +569,9 @@
         },
         {
           "path": "json/personal_tkrworks.json"
+        },
+        {
+          "path": "json/personal_mingaldrichgan.json"
         },
         {
           "path": "json/backtick_z_slash_as_ctrl.json"
--- a/docs/json/matias_ergo_pro.json
+++ b/docs/json/matias_ergo_pro.json
@@ -1,0 +1,19169 @@
+{
+  "title": "Matias Ergo Pro Keyboard",
+  "maintainers": [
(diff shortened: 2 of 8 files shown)


tekezo:
Thanks!

Note:

`docs/json/matias_ergo_pro.json` seems not to be generated from `src/json/matias_ergo_pro.json.rb`.

I reproduced `docs/json/matias_ergo_pro.json` from ruby script and commit it.
https://github.com/pqrs-org/KE-complex_modifications/commit/79efc59fbd3d62d068e6898512b384b4411e1a2b

Please confirm the change.

mingaldrichgan:
> `docs/json/matias_ergo_pro.json` seems not to be generated from `src/json/matias_ergo_pro.json.rb`.
> 
> I reproduced `docs/json/matias_ergo_pro.json` from ruby script and commit it.
> [79efc59](https://github.com/pqrs-org/KE-complex_modifications/commit/79efc59fbd3d62d068e6898512b384b4411e1a2b)

Thanks! It looks like the discrepancy is due to the following expression:
```
Rules.constants.map(&Rules.method(:const_get)).map(&:rule)
```

Our respective versions of Ruby seem to be returning the result of `const_get` in different orders. I expected it to preserve the order in which the modules are defined, but my local builds are inverting the order.

PR #539, opened 2019-09-20
US keyboard with Croatian letters

mightymatth:
This kind of keyboard is available on Ubuntu and I got very familiar with it. It's elegant for programming and also has easy accessible Croatian letters. Check it out, please.


tekezo:
Thanks!

PR #541, opened 2019-09-28
add rules for change shift+insert and shift+fn to command+v

anagura:


tekezo:
Thanks!

anagura:
thanks for merging!

PR #545, opened 2019-10-05
Disabled lazy option

zawataki:
If the lazy option is true, clicking a link while pressing the Command key on Google Chrome doesn't open the link in a new tab.

Additionally, the lazy option was needed to enable key mapping change when the key in "from" and another key are pressed at the same time only.
But it is not needed now because "to_if_alone" works instead of that.

### DIFF
--- a/docs/json/japanese.json
+++ b/docs/json/japanese.json
@@ -19,8 +19,7 @@
           },
           "to": [
             {
-              "key_code": "left_command",
-              "lazy": true
+              "key_code": "left_command"
             }
           ],
           "to_if_held_down": [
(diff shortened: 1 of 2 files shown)


zawataki:
Hi @tekezo, could you please review this PR if you have time?

tekezo:
I prefer using `to_if_held_down` instead of removing `lazy`.

I merged the change into master with `to_if_held_down` modification.
https://github.com/pqrs-org/KE-complex_modifications/commit/25932655418cd0e9e1ae9998fce14e5ee148bbba

Is your issue also solved at master branch?

zawataki:
Thanks. I will check it.

zawataki:
@tekezo I confirmed your change solved this issue, so close this.

tekezo:
Thank you for your confirmation!

PR #549, opened 2019-10-10
Add 'w' to list of left-handed keys

dougjohnston:
The 'w' key was missing from the list of left-handed keys to block when pressing a left-handed modifier.


tekezo:
Thanks!

PR #555, opened 2019-10-23
Added autoshift rules

jhelvy:
Engages "shift + key" by holding the key, and "key" by tapping it. Only applies to alpha & number keys, including tilde and backslash.

A KE implementation of https://github.com/qmk/qmk_firmware/blob/master/docs/feature_auto_shift.md

In response to this issue: 
https://github.com/tekezo/Karabiner-Elements/issues/1395

### DIFF
--- a/docs/groups.json
+++ b/docs/groups.json
@@ -650,6 +650,9 @@
         },
         {
           "path": "json/workaround_broken_sw2tabesc.json"
+        },
+        {
+          "path": "json/autoshift.json"
         }
       ]
     }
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

hqman:
Autoshift is a powerful feature, thank you for that. However, I've encountered an issue where the long-press delay is too long. How can I reduce the delay time?

```
"parameters":{
"basic.to_if_held_down_duration_milliseconds": 100,
                    "basic.to_if_held_down_threshold_milliseconds": 200,
}

```

I try this config but can't work.

PR #556, opened 2019-10-26
Fixed the u (home) and o (end) keys on TouchCursor Mode

donniebreve:
The default setting (to: "home", "end") did not work in most applications. This PR changes the output to command + left/right arrow.


tekezo:
Thanks!

PR #568, opened 2019-11-14
remove f18->spacebar, escape+1,2->brightness control, add spaecbar ->…

tekezo:
… left_shift @ personal_tekezo.json

### DIFF
--- a/docs/json/personal_tekezo.json
+++ b/docs/json/personal_tekezo.json
@@ -5,7 +5,7 @@
   ],
   "rules": [
     {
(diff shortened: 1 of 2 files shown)


nikitavoloboev:
Just curious @tekezo, have you ever considered using https://github.com/yqrashawn/GokuRakuJoudo to generate your Karabiner config?

tekezo:
I think it's good to use such generators.

In my case, I sometimes uses new feature which is not included stable releases while Karabiner-Elements developing.
Thus, I'm using a primitive generator in order to make non-standard rules.

PR #570, opened 2019-11-16
Jetbrains on OSX mimics modifier placement of PC

fahall:
When in Jetbrains, swap the modifier keys to allow easier syncing of settings between PC and OSX via Jetbrains account sync or a settings repository.


tekezo:
Thanks!

PR #571, opened 2019-11-16
Spectacle window snapping can be called with fn + wasdf

fahall:
Call Spectacle with fn and the WASD keys to snap window to edge. fn+f makes the current window full-screen. 

(assumes default Spectacle keymap)


tekezo:
Thanks!

PR #573, opened 2019-11-19
update swiss config 

Birate:
I've rewritten the **Swiss shortcut config** with the ruby syntax

### DIFF
--- a/docs/extra_descriptions/swiss_pc_shortcuts.json.html
+++ b/docs/extra_descriptions/swiss_pc_shortcuts.json.html
@@ -1,12 +1,5 @@
 <h2>Swiss PC-Like Shortcuts</h2>
 <p>
-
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

tekezo:
We can change left_option+xxx combination by the following `from_key_with_option` update.
Please confirm if you want.

before:

```ruby
def from_key_with_option(key_code)
  {
    'key_code' => key_code,
    'modifiers' => Karabiner.from_modifiers(['right_option'], ['left_option']),
  }
end
```

after:

```ruby
def from_key_with_option(key_code)
  {
    'key_code' => key_code,
    'modifiers' => Karabiner.from_modifiers(['option'], ['caps_lock']),
  }
end
```

Birate:
@tekezo Looks like a good idea, I won't test it but that should works anyway 👍

PR #578, opened 2019-11-25
Add `Change shift + delete to tilde` rule to `Change delete key`

tomoya:
I tested on the server and install to assets.

image


tekezo:
Thanks!

PR #581, opened 2019-11-26
add capslock-plus mode

BryanHoo:


tekezo:
Thank you for PR!

Your css in capslock-plus.json.html will be applied entire page due to the current extra description html injection method. (e.g., this css changes font of all text.)

So I'll change the injection method to support scoped css, and then I'll merge this PR.
Please wait for a while.

PR #584, opened 2019-11-29
Add 'Exchange Command+Forward-Delete and Option+Forward-Delete'

Fs00:
I've added a new rule that works like _Exchange Command+Backspace and Option+Backspace_, but with Forward-Delete instead of Backspace.


tekezo:
Thanks!

PR #589, opened 2019-12-04
Add Apple Keyboard only sub-rule for `Section sign (§) to Escape` rule

kopach:


kopach:
https://pqrs.org/osx/karabiner/complex_modifications/#section_sign_to_escape

PR #598, opened 2019-12-16
add modifiers_to_arrows complex modification

thoelze1:
I added a complex modification that remaps modifier keys to arrow keys when pressed alone. I followed the instructions on the readme: `make` succeeded and the `index.html` looked good after running `make server`. Thanks!


tekezo:
Thanks!

PR #602, opened 2019-12-22
Roon application specific key mappings for controlling volume

torarnek:


torarnek:
If you decide to accept this pull request, then I think it is fine to squash these commits.

tekezo:
Thanks!

PR #604, opened 2019-12-23
Fix the mutiple modifier to be mandatory bug & add optional ⌘L

ryuheechul:
I've actually managed to mess it up with this pull request, https://github.com/pqrs-org/KE-complex_modifications/pull/600. (I'm sorry 😅)
The PR was suppose to make Command_L optional but instead it made both L and R mandatory at the same time.
So this PR fix the issue and make Command_L optional, too.

### DIFF
--- a/public/json/cmd_return.json
+++ b/public/json/cmd_return.json
@@ -10,8 +10,7 @@
             "key_code": "return_or_enter",
             "modifiers": {
               "mandatory": [
-                "left_command",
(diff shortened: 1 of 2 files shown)


tekezo:
Thank you for feedback!
Using `command` at modifiers.mandatory is properly way for this purpose.

ryuheechul:
Now it works as intended. Thanks!!

PR #610, opened 2019-12-29
This will make f14 and f15 the copy and paste keys

bstone100:
This will make f14 and f15 the copy and paste keys so print screen can be the screenshot button.


tekezo:
Thanks!

PR #613, opened 2020-01-10
Map ctrl+delete to option+delete

jonathan-golorry:
I use the rule for switching ctrl+arrows with option+arrows so that my ctrl+left/right go to the beginning/end of the current word. This pull request adds rules for doing the same thing with ctrl+back/forward delete.


tekezo:
Thanks!

PR #614, opened 2020-01-10
print_screen to left_command+left_shift+4

0x15e:
Added mapping from print_screen to left_command+left_shift+4 for using PC print screen button to capture screen regions. Tested with macOS Mojave 10.14.6 and KE 12.8.0.


tekezo:
Thanks!

PR #628, opened 2020-02-17
Right option as an EMOJI 😊 key

Rastafabisch:
I kinda missed the touchbar emoji keyboard when using my MacBook Pro in clamshell mode with my eGPU.

Thus I created this inspired by Microsoft.

This transforms the right option key to an emoji key like found on some recent microsoft keyboards.
https://www.reviewgeek.com/thumbcache/0/0/7514b702e9fd8fb1dc73e9b70ca8dbeb/p/uploads/2019/10/98fb5154.jpg


tekezo:
Thanks!

Issue #585, opened 2019-12-01
Config Generator


mikelxc:
It would be nice to have an online complex modifications generator that can generate JSON files based on users' settings so that they can create & import it directly from web.

jhelvy:
I suspect this will be quite difficult to build, but there are existing tools that can dramatically simplify the process of creating the JSON file for complex modifications. For example, check out GokuRakuJoudo: https://github.com/yqrashawn/GokuRakuJoudo

jimpsson:
I have thought of doing exactly this using mainly Vue.js as the frontend framework, if someone would like to create a project together, then I'm interested in putting in some of my time to actually create this, it's totally possible and ideas is nothing I lack, just don't want to create something like this alone...

mikelxc:
Cool, maybe we can do it together

genesy:
I made one today but it's unfinished if you guys want to check it out https://genesy.github.io/karabiner-complex-rules-generator/

PR #640, opened 2020-02-27
Adds Edit JSON dropdown

genesy:
I'm currently making a free open sourced webapp made with react to help edit complex modifications. Is it ok if we add a link from the available repo so people can easily edit them? Also i haven't used vuejs before so i'm not sure if I should be adding the generated/modified files to this pull request.

Knowing people will be using my webapp will also motivate me to fix issues and update it. 

Also solves https://github.com/pqrs-org/KE-complex_modifications/issues/585


tekezo:
Great, Thanks!

> I should be adding the generated/modified files to this pull request.
Thank you for your careful work.
I prefer generated files are excluded from PR like your commit.

PR #643, opened 2020-03-09
Add Ergoemacs mode

marlonrichert:
Ergonomic anti-RSI keybindings for text editing, based on the US Qwerty version of [ergoemacs-mode](http://ergoemacs.github.io/key-setup.html)


tekezo:
Thanks!

PR #645, opened 2020-03-11
Improve navigation in terminal apps

marlonrichert:
* Add support for keyboard with dedicated `delete_forward` key. (Does not emit `fn`.)
* Add info about zsh support.


tekezo:
Thanks!

PR #646, opened 2020-03-14
Add rule : Shift Home and Shift End

satomikko94:
I'd like to add these new Rules 
- Shift Home to Shift Command Left
- Shift End to Shift Command Right

Can you check it?


tekezo:
Thanks!

PR #652, opened 2020-03-22
add left_option + hjkl to arrow config

heychenfq:
I add a rule map left_option + hjkl to arrow keys for myselft. hope can published to public website, thank you ~


tekezo:
Thanks!

PR #674, opened 2020-04-29
Modify Autoshift to split into 3 categories

harshasrisri:
- Autoshift Alphabets only
- Autoshift Numbers on number row only
- Autoshift Special Character keys only


tekezo:
Thanks!

PR #675, opened 2020-05-01
Change modifier key on emacs binding @itmammoth (personal)

itmammoth:
Updated my personal definition.
Thx!


itmammoth:
On my own...

PR #679, opened 2020-05-05
add bash.json for helpful rules for bash command prompt

sdondley:
couple of simple rules for easing the pain of typing paths to the home directory


tekezo:
Thanks!

PR #681, opened 2020-05-06
Swap >/< and °/§

dbirdz:
In european keyboards at least, the key next to 1 and the key next to LShift are swapped in rdp


tekezo:
Thanks!

PR #690, opened 2020-05-15
complex_modification: f2 to cut, f3 to copy, f4 to paste

sinoridha:
Make it (http://xahlee.info/linux/linux_set_F2_F3_F4_to_cut_copy_paste.html) happen on mac.


tekezo:
Thanks!

PR #692, opened 2020-05-16
added stadia_cmd complex mod

sqbush:
Added complex mod that will allow the use of CMD for ALT in google chromes apps.  Specifically designed for Stadia.


tekezo:
Thanks!

PR #694, opened 2020-05-17
Not to use lazy for eisuu/kana to option rule

ypresto:
Option is modifier key, so we can immediately send without waiting for held down.
This removes latency in mouse-with-modifier operations and context menu preview.

Also, with this Karabiner does not send eisuu/kana for slightly longer press, so context menu won't close after previewing option key actions.


tekezo:
Thanks!

PR #699, opened 2020-05-31
lock screen with eject

ldanilek:
I like using the "Eject" key as screen lock (Control+Command+q).

Also, I use the Dvorak input source, so mapping to X instead of Q works better for me and other users of Dvorak.


tekezo:
Thanks!

PR #702, opened 2020-06-03
added Esc to en_US/ABC keyboard for Vim user with non-English based users

ajkj:
added Esc to en_US keyboard switch for Vim user with non-alphabet country(ex.CJK).


tekezo:
Thanks!

PR #708, opened 2020-06-19
Added rule Keychron K2

nbir:
This rule modifies the Keychron K2 keyboard layout to more closely resemble an Apple keyboard.

- Changes right_control to right_option
- Swaps right_command and fn
- Disable page_up, page_down, home, and end


tekezo:
Thanks!

PR #709, opened 2020-06-20
Fixed Keychron K2 rule to work with any modifier key

nbir:
Previously, the rules would behave erratically based on the order in which key combinations are pressed.
e.g. shift+right_option+down_arrow would behave differently from right_option+shift+down_arrow.


tekezo:
Thanks!

PR #710, opened 2020-06-22
Windows shortcuts on macos

sebnil:
This commits adds two keymaps. One to make macOS with an external keyboard behave like it would on Windows, and one keymap that is specific for making Swedish keyboards behave like they would on Windows.


tekezo:
Thanks!

PR #711, opened 2020-06-24
Add a rule: Vimode_with_smart_caps 

LanceSuen:
This rule provides three features without conflicts:
1. Post ESCAPE if CAPSLOCK is pressed.
2. Post CAPSLOCK if ESCAPE is pressed.
3. Map CAPSLOCK + HJKL to arrow keys.


tekezo:
Thanks!

PR #715, opened 2020-07-04
Add CapsLock+Arrows tabs navigation

kevincittadini:
This Karabiner Complex Modification that will allow you to navigate through tabs using Caps Lock + Arrows instead of LCommand+LAlt+Any arrow


tekezo:
Thanks!

PR #718, opened 2020-07-11
Fixed Command_R only, send KANA/EISUU (toggle)

nauleyco:
Previously, the Command_R key also worked as the Command key, but I fixed it to simply send KANA/EISUU.


tekezo:
Is the description of the rule correct?
https://github.com/pqrs-org/KE-complex_modifications/blob/caf2bca2b1f8bd45838dff3afa280f247a963d1c/public/json/japanese.json#L300

I think it should be provided as new rule.

nauleyco:
@tekezo The description is correct. I fixed behavior for it.

tekezo:
I think the description should be like `右コマンドキーを、英数・かなのトグルに変更する`.
Isn't it?

nauleyco:
@tekezo Fair enough. I'll change it.

nauleyco:
@tekezo done. Could you please review and marge it?

tekezo:
Thanks!

PR #723, opened 2020-07-24
Add NuPhy NuType-F1 keyboard changes

alias-mac:
More information about this device:
https://nuphy.com

This change allows the keyboard to work more like expected experience.
In kickstarter page:
https://www.kickstarter.com/projects/nuphy/nutype-revolutionizing-the-laptop-typing-experience/description
this was the original intent, but then modified before shipping the
units to production:
> For example, you can press 'FN' + '~' to substitute the use of 'ESC' button.

When writting markdown it is normal to use '`' often to add some code
blocks. Having to press 'FN' to get the normal key to work doesn't seem
natural and doesn't work like other keyboards that have the 'ESC' button
in a different location (as expected - like new MacBook Pro 16").

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -747,6 +747,9 @@
         },
         {
           "path": "json/hub16-launchapps.json"
+        },
+        {
+          "path": "json/NuType-F1.json"
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


alias-mac:
The reason why I added this config was mostly because this keyboard was already mentioned in Karabiner-Elements repository:
https://github.com/pqrs-org/Karabiner-Elements/issues/2139

I believe that more folks are using this keyboard with KE and might find this configuration useful.

tekezo:
Thanks!

PR #724, opened 2020-07-24
Add "m" to block_one_handed_combos.json

harmtemolder:
The "m" wasn't blocked with right-handed shift. Now it is.


tekezo:
Thank you for PR!

The `public/json/block_one_handed_combos.json` is generated by `src/json/block_one_handed_combos.json.erb`.
Could you change the erb file, too? 

(You can generate the json file by `make` command
https://github.com/pqrs-org/KE-complex_modifications#add-rules )

harmtemolder:
I've never worked with erb files before, hope I did it correctly. I found out the right-handed numbers were also missing, so added those too. Then ran `make` for a new json file.

harmtemolder:
PS Just curious: for this particular case, would it be possible to write a loop inside the erb file to iterate over an array `["shift", "control", etc]` to easily generate the json for all four modifiers?

tekezo:
Thanks!

You can use loop as follows:

```erb
<% ["a", "b", "c"].each do |key| %>
        "<%= key %>": "<%= key.upcase %>",
<% end %>
```

harmtemolder:
@tekezo I tried a loop, but it does not work. Can you tell me what I'm doing wrong, please: https://github.com/harmtemolder/KE-complex_modifications/blob/master/src/json/block_one_handed_combos.json.erb

tekezo:
erb does not allow nested `<%=` like `<%= from("<%= key %>"`.

The correct erb is here:
https://gist.github.com/tekezo/c7dece35846b54a30d6ac720a0c02da6

But in this case, the ruby code might be more suitable.
Rename `block_one_handed_combos.json.erb` to `block_one_handed_combos.json.rb` and put the following contents.
https://gist.github.com/tekezo/f41c281d54158a8e9899477d455a1c18

harmtemolder:
> erb does not allow nested `<%=` like `<%= from("<%= key %>"`.

That's what I was afraid of, but your solution makes sense. Thanks for your help.

The reason I dove into this one again is that I blocked the <kbd>6</kbd>, which probably shouldn't be blocked for either left or right. I'll open up a new PR with that fix.

PR #725, opened 2020-07-26
Add Archiss Quattro TKL JIS layout

kuronekomichael:
More information about this device
https://www.archisite.co.jp/products/archiss/quattro/tkl-jp/

Quattro TKL JIS is a Japanese keyboard device for PC.
This keyboard has a unique key layout that the left and right of the space key replace to mouse buttons.
But the Japanese general key layout for macOS has international keys  "Kana" and "Eisu" at the left and right of the space key. These necessary keys have been stolen, so that keyboard is very difficult to use as Japanese macOS users.

This rule recovers and replaces the international keys for Quattro TKL JIS.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -750,6 +750,9 @@
         },
         {
           "path": "json/NuType-F1.json"
+        },
+        {
+          "path": "json/quattro_tkl_for_mac_jis.json"
         }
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #728, opened 2020-08-01
Unblocked “6” key for both left and right

harmtemolder:
I blocked <kbd>6</kbd> in the previous fix, which should in fact be unblocked for both left and right (like <kbd>B</kbd>). Sorry about that.

Credits to @tekezo for helping me rewrite the ERB into loops for more clarity.


tekezo:
Thanks!

PR #732, opened 2020-08-14
added rule for MouseButton 5 to Ctrl + MouseButton 1

aleksvatman:
added rule for MouseButton 5 to Ctrl + MouseButton 1


tekezo:
Thanks!

PR #735, opened 2020-08-21
Modify @minacle personalisations

minacle:


minacle:
I think that reason of check failure is because of my merge was too fast. It should has no problem.

PR #740, opened 2020-09-01
Add Vim Mode Plus

harmtemolder:
This is the ultimate Vim Mode for Karabiner: NORMAL & VISUAL modes, deleting, yanking, changing, anything thats feasible really.

PS Also ordered the Modifier Keys group alphabetically.

### DIFF
--- a/public/extra_descriptions/vim_mode_plus.json.html
+++ b/public/extra_descriptions/vim_mode_plus.json.html
@@ -1,0 +1,101 @@
+<link rel="stylesheet" href="../../vendor/css/bootstrap.min.css" />
+
+<p style="margin-top: 20px; font-weight: bold">This mod takes your Mac even closer to Vim than all others with both NORMAL and VI modes, yanking and deleting, and more.</p>
+
+<p style="margin-top: 20px; font-weight: bold">
+  Switching modes
+</p>
+
+<table class="table">
+  <tr>
+    <td>Caps Lock</td>
+    <td>Enter NORMAL mode</td>
+  </tr>
+  <tr>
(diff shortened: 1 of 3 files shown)


MuhammedZakir:
Nice! 👍 ❤️  Caps lock key is used to switch input source if enabled in keyboard settings. So, maybe change it to modifier+key? Just a suggestion!

harmtemolder:
Thank you, also for the tip. Is that something a lot of people use <kbd>caps lock</kbd> for?

Also, I got an error on the build of my last commit: https://github.com/pqrs-org/KE-complex_modifications/pull/740/commits. Should I be worried about that?

MuhammedZakir:
> Thank you, also for the tip. Is that something a lot of people use caps lock for?

You're welcome! :-) No! As you can only use it to switch to/from last used Latin input source, most probably, only people dealing with Asian languages use it.

> Also, I got an error on the build of my last commit: https://github.com/pqrs-org/KE-complex_modifications/pull/740/commits. Should I be worried about that? Or is that just because I deleted my fork after merging this pull request?

It happened because you deleted that branch before the build was run -- https://github.com/pqrs-org/KE-complex_modifications/actions/runs/234927062. Also, GitHub now shows `unknown repository` for these commits, as seen at the top. This is not an issue here, but in some cases, it could cause problems. In case you encounter a similar issue in future, a workaround can be found here -- https://github.com/isaacs/github/issues/168#issuecomment-374201226.

---

If you are going to create a new PR, please add a link to your Hammerspoon config in description. That will be helpful.

PR #741, opened 2020-09-05
Adding Hungarian PC layout

Tamas-Biro1:
Replaces the special character positions in Hungarian MacOS keyboard layout to the Hungarian ISO PC keyboard layout style.

examples:
```
RIGHTALT+X gives '#'  instead of pressing SHIFT+OPTION+3
RIGHTALT+Y gives '<'  instead of pressing SHIFT+OPTION+Y
RIGHTALT+Q gives '\'  instead of '@'
```
plus, adds support for special national characters like: ä ě š č

For more details and key combinations, see the PC ISO Hungarian layout here: https://en.wikipedia.org/wiki/QWERTZ#/media/File:KB_Hungary.svg

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -379,6 +379,9 @@
           "extra_description_path": "extra_descriptions/german_programming.json.html"
         },
         {
+          "path": "json/hungarian_pc.json"
+        },
+        {
           "path": "json/us-cro.json"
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #742, opened 2020-09-06
Create durgod-iso-uk-pt-extras-macos.json

ruionwriting:
British Keyboard (DURGOD Taurus K320) on MacBook Pro with pt-PT special chars support. These set of rules can make any Windows UK layout to work very well for any MacOS machine.


ruionwriting:
I'll update this PR shortly and I'll split the rules in smaller groups. It makes a lot of sense to do so.

tekezo:
Thanks!

PR #743, opened 2020-09-06
added eject to screenshot rule

pauloamgomes:
Simple rule that maps Eject key from magic keyboard to shift-cmd-5 (screenshot)


tekezo:
Thanks!

PR #744, opened 2020-09-08
Add keychron K6 f1 helper shortcuts

sourabhv:
Keychron has <code>\`</code> and `~` on Escape key and to use them one needs to use their f1 helper key 

- `f1` + `esc` = `grave`
- `f1` + `shift` + `esc` = `tilde`
- `f1`+ `shift` + `command` + `esc` = `command` + `esc` (to switch windows of same app)

This commit adds shortcuts for those

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -715,6 +715,9 @@
         },
         {
           "path": "json/quattro_tkl_for_mac_jis.json"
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #745, opened 2020-09-09
Add Trader Workstation copy-paste fix

FelikZ:
On mac Trader Workstation still uses `ctrl+` operations instead of `cmd+`.
It swaps ctrl <=> cmd while keeping cmd+space and cmd+tab work as expected.


tekezo:
Thanks!

PR #747, opened 2020-09-13
Neo2 layer 4 (mod4) lock

Matzkowsky:
This PR adds layer 4 (mod4) lock functionality for Neo2 keyboard layout.

Instructions: simply **press mod4** modifiers such as "grave_accent_and_tilde" or "right_command" **alone** to lock or unlock layer 4 of this nice keyboard layout.

The additional functionality is automatically included in presets:
- "Neo2 mod 3 and 4 keys (Apple keyboard)"
- "Neo2 mod 3 and 4 keys (Windows keyboard)"

@jgosmann please review and comment

### DIFF
--- a/src/json/neo2.json.erb
+++ b/src/json/neo2.json.erb
@@ -4,7 +4,9 @@
         <% def set_mod_4(b)
             { "set_variable" => { "name" => "neo2_mod_4", "value" => b } }
         end
-        if_mod_4 = { "type" => "variable_if", "name" => "neo2_mod_4", "value" => 1 }
+        $if_mod_4_on         = { "type" => "variable_unless", "name" => "neo2_mod_4", "value" => 0 }
+        $if_mod_4_locked     = { "type" => "variable_if",     "name" => "neo2_mod_4", "value" => 2 }
+        $if_mod_4_not_locked = { "type" => "variable_unless", "name" => "neo2_mod_4", "value" => 2 }
         def neo2_modifiers(mod_3_key, mod_4_key)
             JSON.generate(
                 each_key(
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

src/json/neo2.json.erb
```diff
@@ -1,10 +1,18 @@
...
-        <% def set_mod_4(b)
-            { "set_variable" => { "name" => "neo2_mod_4", "value" => b } }
+        <%
+        # halt defaults to 'false'
+        def set_mod_4(value, halt = false)
+            if (halt)
+                { "set_variable" => { "name" => "neo2_mod_4", "value" => value }, "halt" => true }
+            else
+                { "set_variable" => { "name" => "neo2_mod_4", "value" => value } }
+            end
```
jgosmann, line 11: I'm not super familiar with ruby, but is this `if` required or could one just use
```
{ "set_variable" => { "name" => "neo2_mod_4", "value" => value }, "halt" => halt }
```
?

jgosmann: Also is the "halt" variable actually used anywhere? It seems, the halt status is handled via "value" being `2`?

jgosmann (commented):
Thank you for the contribution! Just had time to look over the changes and had a question regarding potential simplifications of the additions, but looks good overall.

src/json/neo2.json.erb
MuhammedZakir: `halt` is not "variable" like `value`, but a "flag" like `lazy` and `repeat`. It is used to cancel/stop events. Your changes will add `"halt": false` to config, but generally, you don't add those flags unless you need it. So `if` is better in this case.

See https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/to/halt/.

christophebeling:
Is there some opportunity to deactivate this feature? I'm pretty much used to the none locking layer 4.

Matzkowsky:
> Is there some opportunity to deactivate this feature? I'm pretty much used to the none locking layer 4.

Hi @christophebeling . I didn't found a way to make this option configurable as for example with "Toggle caps_lock by pressing left_shift + right_shift at the same time". However I understand that some users don't like mod4 lock by double press mod4 key. One possible approach would be to have have 2 "Apple keyboard" layouts, 1 with, 1 without lock. Would that be an acceptable solution?

christophebeling:
@Matzkowsky For me, it even locks with a single press. The solution with the 2 layouts would be fine though.

david0:
Are these fixes already available via https://ke-complex-modifications.pqrs.org/? 
I deleted all my rules and imported them via "Import rules from the internet" but I have still the problem of a locking to layer 4 when I press mod4.

jgosmann:
They should be available. Deleting the enabled rules, re-importing the rules, and enabling them again should work ...

Matzkowsky:
I just checked and these changes are not updated in the rules portal. In the meantime you could unpack the attached archive and place the json file under ~/.config/karabiner/assets/complex_modifications/
[neo2.json.zip](https://github.com/pqrs-org/KE-complex_modifications/files/5601956/neo2.json.zip)

jgosmann:
Indeed, the [deployment failed](https://github.com/pqrs-org/KE-complex_modifications/runs/1438517877). Not sure why or I am even able to do anything about that.

jgosmann:
Opened #781 to get this fixed.

jgosmann:
#781 got merged, seems like the changes are now available in the portal.

PR #748, opened 2020-09-14
Add RShift+ESC to tilde for HHKB Professional 2

wuliang142857:
In the HHKB Professional 2 layout, it is very difficult to input the tilde, so map RShift+ESC to the tilde.


tekezo:
Thanks!

PR #749, opened 2020-09-16
Add 'When you type a Shift only, send KANA/EISUU'

akikgt:
Some people (including me) prefer Shift because accidental keyboard shortcut conflicts rarely happen.


tekezo:
Thanks!

PR #751, opened 2020-09-19
add rules map capslock to COMMAND if pressed with other keys and to ESC if pressed only.

phucnsp:


public/groups.json
```diff
@@ -66,8 +66,10 @@
         },
         {
           "path": "json/remap_rshift_esc_to_tilde.json"
-        }
-      ]
+        }, 
+	{
+	"path": "json/capslock2command.json"
+	}]
```
MuhammedZakir, line 72:
```suggestion
       {
          "path": "json/capslock2command.json"
       }
     ]
```

public/json/capslock2command.json
```diff
@@ -0,0 +1,31 @@
...
+          ],
+          "to_if_alone": [
+            {
+              "key_code": "escape"
+            }
+          ]
+        }
+      ]
+    }
+]
```
MuhammedZakir, line 30:
```suggestion
  ]
```

public/json/capslock2command.json
```diff
@@ -0,0 +1,31 @@
...
+            "modifiers": {
+              "optional": [
+                "any"
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "left_command"
+            }
```
MuhammedZakir, line 20: It will be better if you add `lazy` modifier flag [1]. Without it, `command` key will be send even when `caps lock` is pressed alone.

[1] https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/to/lazy/

```suggestion
            {
              "key_code": "left_command",
              "lazy": true
            }
```

tekezo:
Thanks!
I merged this PR.

The lazy flag has pros and cons.
I like the lazy flag because it suppresses unnecessary modifier events.
But it is not good affinity with mouse events.
(The modifier might be ignored if the key press and click happen at the same time.)

So, I left the original code.

PR #752, opened 2020-09-20
Added rshift up if alone rule

wongjingping:
This rule would allow the normal usage of shift in combination with other keys, but allows us to use it as an up key when pressed alone. This is inspired by https://youtu.be/XT2ZmuBbpFE?t=94


tekezo:
Thanks!

wongjingping:
Thanks for accepting it :D

*Wong Jing Ping*
+1 (628) 219-8872

PR #758, opened 2020-10-19
Add auto_close_brackets_and_quotes.json

harmtemolder:
I've been running this mod for a while now so I feel it is ready to be published here: It automatically closes brackets and quotes when you type the left one. E.g. typing a <kbd>(</kbd> will automatically be followed by a <kbd>)</kbd>, after which the cursor is placed between the two. Holding <kbd>fn</kbd> negates this behavior, in case you want to type a single <kbd>'</kbd>, for example.


tekezo:
Thanks!

PR #762, opened 2020-10-22
Fix Ultimate MacOS config

### LINKED ISSUE #716: Can't import rules, Ultimate macOS
MacMini, 2012 Late
macOS, Catalina 10.15.5
Karabiner-Elements, 12.10.0

My native language is Korean.
So I used Google Translator.
Please understand even if the words are strange.

I always use it with gratitude.

I use karabiner's ultimate macOS rule.

1. I left my Mac for a few days.
2. With the power off, connect a 2.4ghz USB type receiver. (Because Mac mini cannot boot without a keyboard and mouse connected.)
3. You are back today.

Karabiner is now unable to find the Ultimate macOS rule and cannot add it.
Other rules can still be added.

To solve this problem
1.It was deleted and reinstalled with karabiner uninstaller. Still having problems
2. I also deleted User files from https://karabiner-elements.pqrs.org/docs/help/advanced-topics/installed-files/.
(First, after deleting the karabiner, and then deleting the user files, and reinstalling)

Still I can't add Ultimate macOS.

Any help would be greatly appreciated.

MuhammedZakir:
Fixes #716.

Importing "Ultimate MacOS" config failed due to no `manipulators` in the first rule.
Added an empty `manipulators` to fix it and also changed `import_url`.

### DIFF
--- a/public/json/ultimate_macOS.json
+++ b/public/json/ultimate_macOS.json
@@ -2,10 +2,11 @@
     "title": "ultimate macOS",
     "maintainers": ["suliveevil"],
     "homepage": "https://github.com/suliveevil/Capslock",
-    "import_url": "karabiner://karabiner/assets/complex_modifications/import?url=https://raw.githubusercontent.com/suliveevil/Capslock/master/mac/ultimate_macOS.json",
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

PR #763, opened 2020-10-22
added carpalx  Q*MLW* usa layouts

0-vortex:
trying out the [carpalx q\*mlw\*](http://mkweb.bcgsc.ca/carpalx/?full_optimization) layout I found it really hard to generate KE json from existing qwerty and colemak configurations so here's an easier to use baseline for them


tekezo:
Thanks!

PR #765, opened 2020-10-25
Minimal dual keys (Modify LEFT SHIFT, LEFT CMD, RIGHT CMD)

yeonsh:
To minimize finger movement this modification changes following key when pressed alone.

* Left SHIFT -> Enter
* Left CMD -> Backspace
* Right CMD -> Enter


tekezo:
Thanks!

PR #772, opened 2020-11-07
Add a rule to switch English and Japanese with the Option key

s4na:
Added new rules for Japanese users.

By using the option key, you can switch between English and Japanese.

---
Best regards,
s4na


tekezo:
Thanks!

PR #777, opened 2020-11-15
Neo2: toggle mod4 lock by pressing left & right mod4 key simultaneously (#747 follow-up)

Matzkowsky:
This is a follow up from github.com/pqrs-org/KE-complex_modifications/pull/747. It allows to toggle mod4 (level4) lock by pressing left & right mod4 keys simultaneously. 
Previously this mode toggle was implemented by just hitting a single mod4 key alone. As some community members weren't happy with this solution I changed it to an implementation which behaves similar to the Linux layout for consistency.
Now the implementation works as follows:

- Rule "Neo2 mod 3 and 4 keys (Apple keyboard).": keys `grave_accent_and_tilde` + `right_command` toggle mod4
- Rule "Neo2 mod 3 and 4 keys (Windows keyboard).": keys `non_us_backslash` + `right_command` toggle mod4

@jgosmann, @christophebeling, @neoissue please check. I would love to get your feedback :-)

### DIFF
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -2,7 +2,7 @@
   "title": "Neo2",
   "rules": [
     {
-      "description": "Neo2 mod 3 and 4 keys (Apple keyboard)",
+      "description": "Neo2 mod 3 and 4 keys (Apple keyboard). Toggle mod4 by pressing both mod4 keys simultaneously.",
       "manipulators": [
         {
           "type": "basic",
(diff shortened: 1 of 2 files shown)


jgosmann (changes requested):
Unfortunately, this does not work properly for me. When I enter mod4 lock by pressing both <kbd>mod4</kbd> keys simultaneously, I seem to be still on layer 1 (typing normal letters). However, pressing <kbd>mod4</kbd> + some other key the normal mod4 actions are not happening. Instead the system registers that as unrelated <kbd>Cmd</kbd> + something else shortcuts.

jgosmann (approved):
My mistake, I was still using the old layer 4/6 definitions and not the ones included in this updated rule definition. With that everything seems to work quite fine. 😃 

Thanks for the effort!

PR #781, opened 2020-11-26
Use webfactory/ssh-agent@v0.4.1 in deploy pipeline

jgosmann:
This should fix the problem with the deprecated set-env command [according to the changelog](https://github.com/webfactory/ssh-agent/blob/master/CHANGELOG.md#v041-2020-10-07).

Example of failure: https://github.com/pqrs-org/KE-complex_modifications/runs/1438517877


tekezo:
Thanks!

PR #783, opened 2020-12-05
added keychron k8 mute teleconference software mappings for macOS

itsermo:
These keyboard remapping use the microphone key on the keychron keyboard (which is Fn + Space) to toggle microphone on and off for Cisco Webex, Microsoft Teams and Zoom.

The remap includes a feature to check if a video conference window is in focus, and will perform the remap.  If there is an active conference session, it will put the window into focus and toggle the mute feature.

If there are is no active conference software running, it will do nothing.


tekezo:
Thanks!

PR #784, opened 2020-12-06
right hand frogpad one-handed key remap

clarkm:
This is a remap for the now-dead frogpad hardware one handed keyboard. It is for the right hand. I am going to remap the left handed version soon. 
I put a detailed description in the extra_descriptions dir.
Thank you!


0-vortex:
LGTM, thanks !

PR #787, opened 2020-12-08
Modified version of german_pc_shortcuts 

wollew:
Make the pipe key work on a Dell PC keyboard which sends non_us_backslash keycode on the key with the pipe symbol


tekezo:
Thanks!

PR #790, opened 2020-12-10
frogpad one handed keyboard left hand remap

clarkm:
I did a PR for the right hand last week. This is for the left hand. Thanks!


MuhammedZakir:
Ignore that revert commit. That wasn't intentional and I have already deleted the branch.

PR #791, opened 2020-12-14
Add support for Microsoft Ergonomic Desktop

LunaticoCR:
This adds support for the Microsoft Ergonomic Desktop.
https://www.microsoft.com/en-ie/p/microsoft-ergonomic-desktop/926gk79b71v9


MuhammedZakir:
Why is same key added more than once in `from.simultaneous`? AFAIK, adding a key more than once in `from.simultaneous` do nothing because you can not release a key before other keys in `simultaneous` are pressed.

https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/from/simultaneous/:
> 
> #### Note
> 
> There are some cases `simultaneous` does not modify events.
> 
> * `simultaneous` does not modify events if any `from` events are released before all `from` events are pressed.
> 
> _--snip--_

LunaticoCR:
After analysing the special keys with the EventViewer what I found Microsoft does is that a key press of one of these special keys will send a sequence of keys and using simultaneous seems to work perfectly. I've been using it for a few weeks with no issues.

MuhammedZakir:
> After analysing the special keys with the EventViewer what I found Microsoft does is that a key press of one of these special keys will send a sequence of keys and using simultaneous seems to work perfectly. I've been using it for a few weeks with no issues.

Merged. :-)

MuhammedZakir:
Thinking about it again, I am not sure if duplicate keys in `from.simultaneous` makes any difference. Can you please tell what happens if duplicated keys are removed from `from.simultaneous`. Have you tested it?

LunaticoCR:
> Thinking about it again, I am not sure if duplicate keys in from.simultaneous makes any difference. Can you please tell what happens if duplicated keys are removed from from.simultaneous. Have you tested it?

I'm not sure I understand. What exactly would you like me to try?

MuhammedZakir:
> > Thinking about it again, I am not sure if duplicate keys in from.simultaneous makes any difference. Can you please tell what happens if duplicated keys are removed from from.simultaneous. Have you tested it?
> 
> I'm not sure I understand. What exactly would you like me to try?

Remove all duplicate `key_code`s from `from.simultaneous` and try if it's still working. E.g. https://github.com/LunaticoCR/KE-complex_modifications/blob/4fbf18d2436b9fa03cfb5681705dc3a1a269d8a3/public/json/Microsoft_Ergonomic_Desktop.json#L10-L41

```diff
            "simultaneous": [
              {
                "key_code": "left_command"
              },
              {
                "key_code": "left_option"
              },
              {
                "key_code": "left_shift"
              },
              {
                "key_code": "left_control"
              },
              {
                "key_code": "spacebar"
-             },
+             }
-             {
-               "key_code": "spacebar"
-             },
-             {
-               "key_code": "left_control"
-             },
-             {
-               "key_code": "left_shift"
-             },
-             {
-               "key_code": "left_option"
-             },
-             {
-               "key_code": "left_command"
-             }
            ]
```

LunaticoCR:
I just tested it and you're right, it still works.

The reason why I did it like this is because that's how it showed on the viewer.

image

MuhammedZakir:
> I just tested it and you're right, it still works.
> 
> The reason why I did it like this is because that's how it showed on the viewer.
> 
> `image

You only need to mention same key one in `from` event. However, you may mention same key more than once in `to` event.

All key presses have two states, key down and key up. That's why those keys are shown twice. You can see the event type in the leftmost column. Usually, you don't need to care about the key state. But, sometimes, you need to check these to verify whether your config is correct. For example, to see if order of pressed keys or released keys, is in the expected order.

LunaticoCR:
> You only need to mention same key one in from event. However, you may mention same key more than once in to event.
All key presses have two states, key down and key up. That's why those keys are shown twice. You can see the event type in the leftmost column. Usually, you don't need to care about the key state. But, sometimes, you need to check these to verify whether your config is correct. For example, to see if order of pressed keys or released keys, is in the expected order.
P.S. One more (obvious) thing to note is that, the order of pressed keys and released keys will change depending on the order one presses and releases keys. Even order of key presses of same person can change if he/she uses a different keyboard. So, if you plan to share a config, you should keep the use of "order of key presses" in it to absolute minimum.

Understood. I see you already removed the duplicated so thanks for that! I tested the file and it wasn't loading in Karabiner, an issue with json syntax. Here's a PR for a quick fix: https://github.com/pqrs-org/KE-complex_modifications/pull/801

MuhammedZakir:
Thanks for PR! Really appreciate it! :-) I have posted a comment there. Let's continue there. :-)

PR #798, opened 2021-01-05
Molleweide

molleweide:
Hi, 

This is a work in progress. I am trying to create dual keys on the home row
for modifier keys. This is because on my QMK keyboard I have all mod keys layed out as dual keys on the home row. This feels amazining because I barely have to move my hands at all. BUT this is not working as expected for the pinky-control-dual-key here. I put control when held down on pinkys (a/;) but the control event seems to be triggered on tap as well as when held down. hmmm...

I am also a beginner when it comes to making PRs so I might be doing things in the wrong order..


molleweide:
i have created an issue for my problem with CTRL dual keys,
cmd, shift, and alt works great, but ctrl is not working as it should.
it seems that the ctrl even is being sent also on tap which makes it unusable

PR #800, opened 2021-01-05
Fixed wrong case in BundleID

thomascope:
This did not work for me as bundle_identifiers are case sensitive. My VNC viewer BundleID is all lower case. I don't know when this changed so have modified the code to search for either case.

### DIFF
--- a/public/json/vnc_swap_left_cmd_ctrl.json
+++ b/public/json/vnc_swap_left_cmd_ctrl.json
@@ -16,6 +16,7 @@
             {
               "type": "frontmost_application_if",
               "bundle_identifiers": [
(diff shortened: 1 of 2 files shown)


MuhammedZakir:
Isn't it better to use `^com\\.realvnc\\.(vncv|VNCV)iewer` as bundle identifier regex?

thomascope:
I think either version works. I like mine as being more transparent to end users, but yours is a few bytes fewer! The main thing is that yours and mine work, but the master does not!

MuhammedZakir:
You're right! I will go with being more transparent to end users. :-)

PR #801, opened 2021-01-11
Fix json syntax for Microsoft Ergonomic Desktop

LunaticoCR:
Karabiner wasn't loading this complex modifications file because of a json syntax error.


MuhammedZakir:
I have one more suggestion: change `any` in `from.modifiers.optional` to `caps_lock`. I think `any` may create trouble/conflict as you are already using modifiers in `from.simultaneous`. I think the onnly modifier that need to be added in optional modifiers is `caps_lock`, so that you can use these config when caps lock is enabled. What do you think?

LunaticoCR:
> I have one more suggestion: change `any` in `from.modifiers.optional` to `caps_lock`. I think `any` may create trouble/conflict as you are already using modifiers in `from.simultaneous`. I think the only modifier that needs to be added in optional modifiers is `caps_lock`, so that you can use these config when caps lock is enabled. What do you think?
> 
> P.S. You may also add `fn` if there are situations where you need to use these modifications when `fn` is held down.

Thanks for the suggestion! I'll add the caps_lock to the optional modifiers. The keyboard has no fn key so I'll leave that out.

MuhammedZakir:
> Thanks for the suggestion! I'll add the caps_lock to the optional modifiers. The keyboard has no fn key so I'll leave that out.

Great! :+1:

PR #805, opened 2021-01-23
Add "Mouse keys (xgwang.me)"

xg-wang:
This is my tweak of the "Mouse keys full rev4".
The major change is to press release right ⌘ to toggle mouse keys mode,
rather than holing ⌘.


tekezo:
Thanks!

PR #807, opened 2021-01-24
Config for dvorak with qwerty cmd only applicable to US intput type

andreyshtamburg:
[dvorak_cmd_qwerty_layout.json](https://github.com/pqrs-org/KE-complex_modifications/blob/master/public/json/dvorak_cmd_qwerty_layout.json) works fine when you use only one input type.
In case you used to having more than one input type, it becomes fairly inconvenient.
Proposed config ties changes only to US input type

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -433,6 +433,9 @@
           "path": "json/dvorak_cmd_qwerty_layout.json"
         },
         {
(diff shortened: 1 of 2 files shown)


MuhammedZakir:
Thanks! :+1:

PR #817, opened 2021-02-12
Update windows_shortcuts_on_macos.json, to exclude X11 Macports apps closes #816

### LINKED ISSUE #816: Windows shortcuts on macOS, to exclude X11 Macports apps
Following up on #814 

> `^-C` means "Control - Break" way before it means "copy". 

I cannot miss the ability to do "control - break". But now that the rule needs to be as general as possible and cannot replace all `control` with `left_control` for all rules, I'd propose to exclude all X11 Macports apps from the rules, just like apple Terminal is excluded from the rules too.

I've successfully change it locally. OK I send in the PR?

suntong:
to exclude X11 Macports apps, closes #816

### DIFF
--- a/public/json/windows_shortcuts_on_macos.json
+++ b/public/json/windows_shortcuts_on_macos.json
@@ -1012,7 +1012,7 @@
               "conditions": [
                 {
                   "bundle_identifiers": [
-                                "^org\\.macports\\.X11$",
+                    "^org\\.macports\\.X11$",
                     "^com\\.apple\\.Terminal$",
                     "^com\\.googlecode\\.iterm2$",
                     "^co\\.zeit\\.hyper$",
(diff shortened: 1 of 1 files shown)


MuhammedZakir:
Is this a single app, or are there multiple apps? You used "X11 MacPorts app**s**" (plural).

suntong:
OK. That needs a bit of explain, but basically -- _all X11 have the same bundle identifier_.

Starting from the very beginning

- MacPorts port Linux apps
- all Linux GUI apps use X11
- all such X11 apps from MacPorts have the same bundle identifier (Verified)

MuhammedZakir:
Merged. I am taking your word for it! :wink:

suntong:
Thanks!

Yeah, I double-checked and verified everything 😀

PR #820, opened 2021-02-16
Add 'Touch Bar Lock' complex modification

akrmn:
Hi! I'm a long time user of KE but this is my first time actually writing a modification. I thought I might add it here in case anyone else finds it useful. 

The one thing I'm not sure about is the looong `shell_command`, but I thought that was easier than adding a bash script elsewhere. Please do let me know if there's a better way to achieve that effect.


tekezo:
Thanks!

tekezo:
> looong shell_command

This is the most smart way to do it!

PR #836, opened 2021-03-08
add teams config for KKmoon foot switch

zutherb:
added teams support for the following device:
https://www.amazon.de/gp/product/B00DG0KQBE/ref=ppx_yo_dt_b_asin_title_o02_s00?ie=UTF8&psc=1


tekezo:
Thanks!

PR #837, opened 2021-03-09
Keynote And Illustrator

juonic:
Just sharing the personal setting used in Keynote and Illustrator for the convience of designers. The most importang roles are using the mouse button 4 and 5 for zoom in and zoom out in keynote.


tekezo:
Thanks!

PR #840, opened 2021-03-09
add 'Quickly switch to Chinese input on a JIS keyboard'

PeaShooterR:
Using kana/eisuu key on a JIS keyboard to quickly switch input method between English, Japanese and Chinese.


tekezo:
Thanks!

Issue #843, opened 2021-03-11
Shell commands are not repeated


Rastafabisch:
Hi there!

I just realised this behaviour; the description is pretty self explanatory. Is this intentional?
If it is, could it be changed to default to `"repeat": false` which can be overwritten by `"repeat": true`?

My use-case is sending keyboard shortcuts to an app in the background via an embedded osascript.

PR #844, opened 2021-03-11
Sync Music and System Volume

Rastafabisch:
This one I am particularly proud of. Mainly because I desperately missed an app, I used to use years ago with iTunes, which kept iTunes and system volume in sync. While I just prefer this behaviour generally, it also enables you to control non-system-wide AirPlay audio via the volume keys, which I need quite often.
Still missing is key repeat functionality, as this is apparently not supported for shell commands. Therefor it is intentionally disabled altogether (for the volume keys) to not behave weirdly.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -285,6 +285,12 @@
       "id": "application-specific",
       "files": [
         {
+          "path": "json/Sync_SystemVolume_to_Music.json"
+        },
+        {
+          "path": "json/Sync_SystemVolume_to_iTunes.json"
+        },
+        {
           "path": "json/bash.json"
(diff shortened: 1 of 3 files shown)


Rastafabisch:
[I opened an issue regarding the described behaviour.](https://github.com/pqrs-org/KE-complex_modifications/issues/843)

Rastafabisch:
Sorry for the convoluted process. There were some oversights on my part at first and I did not realised that changes needed to be included into group.json.

tekezo:
Thanks!
Sorry to keep you waiting.

PR #845, opened 2021-03-13
publish cmd/opt->en/ru/uk layout switcher

piranha:
I'm using that for a few years already and just got an idea that it's much better to publish to everyone that just post code on my blog. :)


tekezo:
Thanks!

PR #846, opened 2021-03-15
[ADD] input-source change directly like windows-os for korean

creatorKoo:
Originally mac-os input-source-change triggered from key-up event.
But, windows-os input-source-change triggered key-down event.
So, I add windows-os liked input-source-change function to mac-os for korean.


tekezo:
Thanks!

PR #848, opened 2021-03-28
Map capslock + number keys to F1..F12

rlei:
This rule maps Caps lock and 1,2,3,4...8,9,-,+ to F1,F2,F3,...,F11,F12. Allows shorter finger move distance on 61/68 keyboards :).


tekezo:
Thanks!

PR #858, opened 2021-04-21
Fixed command description for German programming modification

snipem:
Description is wrong it is the key next to "ö" which is "ä" on a German ISO-DE layout.


tekezo:
Thanks!

PR #864, opened 2021-05-11
Keychron K3 Remap £ to # for British Keyboard

fredericrous:
This scripts swaps the pound and the number sign only on Keychron K3 keyboard.
This means I can keep my British layout on my macbook internal keyboard, and use my Keychron K3 as a normal US keyboard


tekezo:
Thanks!

PR #871, opened 2021-05-30
Neo2 layout: Home/end key for your terminal in macOS.

dominikmn:
This rule lets you jump to the beginning and the end of a line in your
terminal directly using ↖ and ↘︎ on layer 4 in macOS.
So far it was not possible to do so via layer 4 in macOS using the standard Karabiner Neo2 rules.


dominikmn:
Instead of putting this rule to a separate JSON, I will merge it into the existing neo2.json and add the authors as reviewers.

PR #873, opened 2021-05-30
Several term keybindings to ease people coming from Linux

jjo:
Add three new entries:
* [1] Swap CMD and Option keys, useful for CMD + f/b/d/./_ shell keybindings
* [2] Add Control_Shift (and [3] Capslock_Shift) + C/V to copy/paste (ala to Linux terms)


tekezo:
Thanks!

PR #874, opened 2021-05-31
Add shortcuts for Apple Mail and Slack

aikinai:
Adding some app-specific shortcuts that are impossible to set with System Preferences.

### Apple Mail
- **⌘-enter to send mail like in lots of other apps and websites**  
I've tried setting this with `defaults write com.apple.mail NSUserKeyEquivalents -dict-add "Send" "@\\U21a9"` and it shows up in the menu correctly, but doesn't actually function, so I think a key remap is required.

### Slack
- **⌘-k to insert or convert to hyperlink like most apps and sites**  
This has no menu item, so it's impossible to set with the OS.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -430,6 +430,12 @@
         },
         {
           "path": "json/KeynoteAndIllustrator.json"
+        },
+        {
+          "path": "json/apple_mail_shortcuts.json"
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

PR #875, opened 2021-05-31
Neo2: Home/end key for your terminal.

dominikmn:
<p>This rule lets you jump to the beginning and the end of a line in your terminal directly using <kbd>↖</kbd> and <kbd>↘︎</kbd> on layer 4. More precisely it maps <kbd>a</kbd> + <kbd>mod4</kbd> to <kbd>Home</kbd> and <kbd>g</kbd> + <kbd>mod4</kbd> to<kbd>End</kbd> in the following terminal apps: macOS Terminal, iTerm2, Hyper, alacritty, kitty.</p>
        <p>Make sure to move this rule <em>above</em> the rule 'Neo2 layer 4' in the Karabiner-Elements preferences.</p>
        <p>In case you use the macOS Terminal: You need to perform an additional configuration step in the macOS Terminal preferences: Go to 'Terminal' -> 'Preferences' -> 'Profiles' -> 'Keyboard' and add the following two entries: Key: ↖ Action: \033OH and Key: ↘︎ Action: \033OF.</p>

### DIFF
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -4164,7 +4164,7 @@
       ]
     },
     {
-      "description": "Neo2 mod 4: Map ↖ to Home and ↘︎ to End in terminal windows.",
+      "description": "Neo2 mod 4: Map ↖ to Home and ↘︎ to End in terminal apps (move this above the rule 'Neo2 layer 4').",
       "manipulators": [
         {
           "type": "basic",
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -4162,6 +4162,83 @@
           ]
         }
       ]
+    },
+    {
+      "description": "Neo2 mod 4: Map ↖ to Home and ↘︎ to End in terminal windows.",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "a",
+            "modifiers": {
+              "optional": [
+                "shift",
+                "caps_lock",
+                "left_option"
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "home"
+            }
+          ],
+          "conditions": [
+            {
+              "type": "frontmost_application_if",
+              "bundle_identifiers": [
+                "^com\\.apple\\.Terminal$",
+                "^com\\.googlecode\\.iterm2$",
+                "^co\\.zeit\\.hyperterm$",
+                "^co\\.zeit\\.hyper$",
+                "^io\\.alacritty$",
+                "^net\\.kovidgoyal\\.kitty$"
+              ]
+            },
+            {
+              "type": "variable_unless",
+              "name": "neo2_mod_4",
+              "value": 0
+            }
+          ]
+        },
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "g",
+            "modifiers": {
+              "optional": [
+                "shift",
+                "caps_lock",
+                "left_option"
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "end"
+            }
+          ],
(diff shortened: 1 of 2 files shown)


dominikmn:
@jgosmann kindly requesting for a review

public/json/neo2.json
```diff
@@ -4162,6 +4162,83 @@
...
+          "type": "basic",
+          "from": {
+            "key_code": "a",
+            "modifiers": {
+              "optional": [
+                "shift",
+                "caps_lock",
+                "left_option"
+              ]
+            }
```
dominikmn, line 4179: I have copied this part from the base rule "Neo2 layer 4". Not sure if it is needed here.

jgosmann: I think the `caps_lock` is required (otherwise the shortcut would not work with activated caps lock). Te `shift` and `left_option` probably make no difference either way. At least for me the additional modifiers caused the shortcuts to be interpreted differently.

jgosmann (changes requested):
The `public/json/neo2.json` is generated from `src/json/neo2.json.erb`, so the latter file needs to be changed too. Otherwise, the new rules will be overwritten the next time `public/json/neo2.json` in regenerated.

I wonder if the requirement to move this in front of the layer 4 rule could be put into the description? It might get too long, but I'm worried that someone trying to use this rule and not aware of this PR won't get it to work with that essential piece of information.

The additional instructions for the macOS Terminal could be added to the README.md of jgosmann/neo2-layout-osx.

dominikmn:
Thanks for the review!

> The `public/json/neo2.json` is generated from `src/json/neo2.json.erb`, so the latter file needs to be changed too. Otherwise, the new rules will be overwritten the next time `public/json/neo2.json` in regenerated.

I hoped to get around it as I'm not familiar with ruby. Do I assume that correctly that the .erb file requires ruby code? Anyway, I will dig into this.

> I wonder if the requirement to move this in front of the layer 4 rule could be put into the description? It might get too long, but I'm worried that someone trying to use this rule and not aware of this PR won't get it to work with that essential piece of information.

Done. I share the same concern. I did not do it _so far_ because I was under the wrong impression that this requirement ("above the layer 4 rule") would also be true for the rule "Toggle mod4 by pressing both mod4 keys simultaneously".

> The additional instructions for the macOS Terminal could be added to the README.md of jgosmann/neo2-layout-osx.

Yes, that would be nice. I have an additonal idea related to that: Meanwhile I have written an extra-description .html file which would be displayed on https://ke-complex-modifications.pqrs.org/ right under the Neo2 rules. I could include it in a separate PR? If you like to have a look already now, you'll find it in a dedicated branch in my fork. Note: To see the html file rendered nicely, one has to start the local server with `make server` (rather than directly opening the html file)

jgosmann:
> Do I assume that correctly that the .erb file requires ruby code?

It is JSON with embedded ruby, if you find the right spot you can probably just insert the the plain JSON.

For your consolation, I don't really know ruby either. ;) I more or less combined snippets from looking at some of the other .erb files.

> I have an additonal idea related to that: Meanwhile I have written an extra-description .html file which would be displayed on https://ke-complex-modifications.pqrs.org/ right under the Neo2 rules.

Cool, I didn't know that was possible. I will have a look, but earliest at the weekend.

jgosmann:
I just had a look at the extra-description .html file and think it looks great! I would definitely welcome a PR for it. :)

dominikmn:
I have extended the .erb file by the new rule.
Successfully tested with `make`. 
Should be ready for merging now as far as I can see

jgosmann (approved):

PR #881, opened 2021-06-15
Create Change Single Key as Combine Keys

ngotuankhoi:


MuhammedZakir:
Pardon, but what if you want to type `c` or `v`. Can you please tell how (or when) you use this modification?

tekezo:
It seems to me that the condition is missing, because we cannot type `c` or `v` with these rules.
Please add `device_if`, `frontmost_application_if`, etc.
I close this PR for now.

PR #888, opened 2021-07-05
Add rules to remap symbols from JIS to US keyboard

halfwhole:
These rules help to remap the symbols on a JIS keyboard to a layout that's much more reminiscent to that of a US keyboard

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -545,6 +545,9 @@
           "path": "json/jis_to_ascii.json"
         },
         {
+          "path": "json/jis_to_us_symbols.json"
+        },
+        {
           "path": "json/german_pc_shortcuts.json"
         },
         {
--- a/public/json/jis_to_us_symbols.json
+++ b/public/json/jis_to_us_symbols.json
@@ -21,6 +21,14 @@
             {
               "key_code": "open_bracket"
             }
+          ],
+          "conditions": [
+            {
+              "type": "keyboard_type_if",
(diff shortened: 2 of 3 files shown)


MuhammedZakir:
In `from.modifiers`, `shift`, `control`, `command` and `option`, all mean either right or left modifier/button. So, instead of writing two manipulators, you only need to write one.

https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/from/modifiers/#list-of-modifiers:

> | Name  | Description |
> | ----- | ----- |
> | command | Either left command or right command is pressed |
> | control | Either left control or right control is pressed |
> | option | Either left option or right option is pressed |
> | shift | Either left shift or right shift is pressed |

---

Do note that this is different from `to.modifiers`. In it, `shit/command/control/option` means left modifier/button. See https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/to/modifiers/.

halfwhole:
Thanks for the advice, have modified it accordingly

MuhammedZakir:
Should have said earlier, but then forgot: can you please add condition to restrict it to `JIS` keyboard type? See https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/conditions/keyboard-type/.

halfwhole:
Yep, works as expected on my device

MuhammedZakir:
Merged. :-)

PR #889, opened 2021-07-06
Maps Eject key to toggle Mute Mic in Microsoft Teams

redmer:
This PR adds a rule to map Eject to an AppleScript that sends ⌘⇧M to MS Teams. It has already proven to useful for noisy environments.


tekezo:
Thanks!

PR #890, opened 2021-07-06
Added right command + v to paste without formatting

jakebox:
Hello,

I've written a small complex modification to make `right command + v` paste without formatting. Pretty useful as it's much easier on the hands and a convenient keybinding for anyone who does a lot of copy-pasting into style-aware documents. Thank you!


tekezo:
Thanks!

PR #892, opened 2021-07-07
Add Hyper Key w/ Caps Lock fallback

slunsford:
There are a bunch of hyper key rules in the directory, but I didn't see any that still function as caps lock when pressed alone. (Most seem to go to Esc—I have an escape key for that.)

### DIFF
--- a/public/json/hyper_key.json
+++ b/public/json/hyper_key.json
@@ -1,5 +1,5 @@
 {
-	"title": "Hyper Key",
+	"title": "Caps Lock → Hyper Key (⌃⌥⇧⌘) (Caps Lock if alone)",
 	"rules": [
 		{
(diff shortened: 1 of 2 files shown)


public/json/hyper_key.json
```diff
@@ -0,0 +1,31 @@
+{
+	"title": "Hyper Key",
```
MuhammedZakir, line 2:
```suggestion
	"title": "Caps Lock → Hyper Key (⌃⌥⇧⌘) (Caps Lock if alone)",
```

"Hyper Key" is too vague for a title, when considering, complex modifications here are meant to be imported by others. :-)

MuhammedZakir (changes requested):

MuhammedZakir:
Merged! :-)

PR #893, opened 2021-07-09
add rule to execute Fish shell autocomplete command with shift + enter

cbejensen:


tekezo:
Thanks!
I updated `groups.json` with this PR.
https://github.com/pqrs-org/KE-complex_modifications/commit/fa8585ec0eed61787612378d671add765745850d

PR #895, opened 2021-07-11
Map Right Command to take screenshot

anirudhsama:
Right command will trigger Left Cmd + Left Shift + 5 to trigger the screenshot UI


tekezo:
Thanks!

PR #896, opened 2021-07-13
Create Orbit_Fusion_with_Button5

FaqT0tum:
With Button5 Active Orbit with mouse pointer in Fusion360 by Autodesk


MuhammedZakir:
Can you please limit this rule to AutoDesk only? See https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/conditions/frontmost-application/.

PR #897, opened 2021-07-14
Add control and option as optional modifiers in JIS to US remappings

halfwhole:
An add-on to previous PR #888, allowing control and option specifically to be optional modifiers in the rules. It's handy when attempting combinations such as shift+alt+;

### DIFF
--- a/public/json/jis_to_us_symbols.json
+++ b/public/json/jis_to_us_symbols.json
@@ -14,7 +14,8 @@
             "modifiers": {
               "optional": [
                 "control",
-                "option"
+                "option",
+                "command"
               ],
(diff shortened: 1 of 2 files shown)


MuhammedZakir:
You may also want to add `command` to `from.optional.modifiers` for non-shifted symbols and `+`. The reason is some apps use them. For example, `command` + `+`/`-` for zooming and tab/pane switching in terminals.

Note: I don't use JIS keyboard and am not familiar with it. This is from a "general" pov. Ignore this if it doesn't fit in this workflow.

halfwhole:
Thanks! I've added `command` as well.

PR #898, opened 2021-07-15
Add Left CTRL -> Hyper key modification

luan-vieira:
A simple rule to remap CTRL to Hyper key since I've re-mapped Caps Lock to CTRL

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -4,6 +4,9 @@
       "name": "Modifier Keys",
       "id": "modifier-keys",
       "files": [
+	{
+	  "path": "json/ctrl_to_hyper.json"
+	},
         {
           "path": "json/hyper_key.json"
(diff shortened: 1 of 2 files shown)


MuhammedZakir:
This won't work if caps lock is on. You may wanna add `caps_lock` to `from.modifiers.optional`<sup>[1]</sup>. If that was intentional, say so, and I will merge this as is. :-)

[1] https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/from/modifiers/#frommodifiersoptional

luan-vieira:
@MuhammedZakir , thanks for the quick feedback and llink to documentation. I implemented your suggestion!

MuhammedZakir:
Merged. :-)

PR #900, opened 2021-07-20
feat: added complex modifier for caps lock

egstad:
1. CAPS LOCK: Converts 'jkl;' to arrow keys

2. CAPS LOCK + SPACEBAR: When held, 'spacebar' becomes 'left_shift' + 'left_opt' for easy text selection


MuhammedZakir:
Merged. :-)

PR #901, opened 2021-07-21
Enhanced rules for Neo2 for mixing keyboards

Kuchenaxt:
The neo2.json rules had two different rule sets for Apple and non-Apple keyboards:

- Neo2 mod 3 and 4 keys (Apple keyboard).
- Neo2 mod 3 and 4 keys (Windows keyboard).

They can’t be active at the same time. Switching between Apple and non-Apple keyboards (e.g. plugging an external keyboard into a Macbook) required changing between rule sets.

This change enhances the rules by checking for the Device ID with device_if.

For example:
```
{
          "type": "basic",
          "from": {
            "simultaneous": [
              {
                "key_code": "grave_accent_and_tilde"
              },
              {
                "key_code": "right_command"
              }
            ]
          },
          "to": [
            {
              "set_variable": {
                "name": "neo2_mod_4",
                "value": 2
              },
              "halt": true
            }
          ],
          "conditions": [
            {
              "type": "variable_unless",
              "name": "neo2_mod_4",
              "value": 2
            },
            {
              "type": "device_if",
              "identifiers": [
                {
                  "vendor_id": 1452,
                  "description": "Apple devices"
                }
              ]
            }
```

The rules are appropriately re-named to:

- Neo2 mod 3 and 4 keys (Apple keyboards). Rule applied to Apple-made keyboards only.
- Neo2 mod 3 and 4 keys (Windows keyboard). Rule applied to all non-Apple keyboards.

Unfortunately this fix doesn't work for third party Mac keyboards (since they don’t have an Apple Device ID). Therefore the "Neo2 mod 3 and 4 keys (Apple keyboard)" rule has been kept and renamed to "Neo2 mod 3 and 4 keys (third party Mac keyboard). Rule applied to all keyboards."

### DIFF
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -1,8 +1,222 @@
 {
   "title": "Neo2",
   "rules": [
+  {
+      "description": "Neo2 mod 3 and 4 keys (Apple keyboards). Rule applied to Apple-made keyboards only.",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "backslash",
+            "modifiers": {
+              "optional": [
+                "any"
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "right_option"
+            }
+          ]
+        },
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "caps_lock",
+            "modifiers": {
+              "optional": [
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

PR #908, opened 2021-08-07
Add binding for '⌥+Tab' to '⌘+Tab'

rustycl0ck:


MuhammedZakir:
As yours is pure json, there is no need for `.erb` file. Also, use `left_command` in place of `left_gui`.

rustycl0ck:
Thanks for the comments. I have removed the `.erb` file and switched from `left_gui` to `left_command`. Also added an extra description file.

public/json/option-tab_to_command-tab.json
```diff
@@ -0,0 +1,33 @@
...
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "tab",
+              "modifiers": [
+                "left_command"
+              ],
+              "lazy": true
```
MuhammedZakir, line 25:
```suggestion
              ]
```

AFAIK, `to.lazy` only works if `from.keycode` is a modifier key.

From https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/to/lazy/:
> If `to.lazy` is `true` and `to.key_code` is modifier flag such as `"key_code": "left_shift"`, the `to.key_code` acts as lazy modifier.

public/extra_descriptions/option-tab_to_command-tab.json.html
```diff
@@ -0,0 +1 @@
+Use '⌥+Tab' to cycle through applications (in addition to the default '⌘+Tab')
```
MuhammedZakir, line 1:
```suggestion
<link rel="stylesheet" href="../../vendor/css/bootstrap.min.css" />

<p>Use '⌥+Tab' to cycle through applications (in addition to the default '⌘+Tab')</p>
```

MuhammedZakir (changes requested):

MuhammedZakir:
I will merge this once those two are solved. Thanks! :-)

public/json/option-tab_to_command-tab.json
rustycl0ck: Thanks for pointing this out. I didn't fully understand the function of `lazy`. I have removed it now.

public/extra_descriptions/option-tab_to_command-tab.json.html
rustycl0ck: Thanks for suggesting this. I have updated it as requested.

MuhammedZakir:
Merged. :-)

PR #911, opened 2021-08-10
Add hhkb-like-arrows-and-page-control

gun0317:
Fn + [;/' to arrows
Fn + l to page up
Fn + . to page down

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -4,9 +4,6 @@
       "name": "Modifier Keys",
       "id": "modifier-keys",
       "files": [
-	{
-	  "path": "json/hhkb-like-arrows-and-page-control.json"
-	},
         {
           "path": "json/ctrl_to_hyper.json"
         },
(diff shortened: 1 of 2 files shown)


public/groups.json
```diff
@@ -4,6 +4,9 @@
       "name": "Modifier Keys",
       "id": "modifier-keys",
       "files": [
+	{
+	  "path": "json/hhkb-like-arrows-and-page-control.json"
+	},
```
MuhammedZakir, line 9: This should be inside "emulation mode" instead of "modifier keys".

MuhammedZakir (changes requested):

public/groups.json
gun0317: Thanks

public/groups.json
```diff
@@ -125,6 +125,9 @@
       "name": "Emulation Modes",
       "id": "emulation-modes",
       "files": [
+	{
+	  "path": "json/hhkb-like-arrows-and-page-control.json"
+	},
```
MuhammedZakir, line 130:
```suggestion
        {
          "path": "json/hhkb-like-arrows-and-page-control.json"
        },
```

MuhammedZakir:
Merged. :-)

PR #917, opened 2021-08-18
Add conditions to avoid apps

Metalliza:
This will add conditions to avoid remotes or virtual machines into For Korean Keyboard rules.


tekezo:
Thanks!

PR #918, opened 2021-08-18
made typematrix bepo distinct and added new one

bartdorlandt:
The current typematrix json is specifically for bepo layout. 
New json added for normal keys as expected by mac (qwerty/dvorak)


tekezo:
Thanks!

PR #919, opened 2021-08-19
Add new rule: left option -> switch to Chinese

yuichkun:
This add the following mod rule:

- When: 
  - left option/alt key is pressed alone
- What: 
  - switch the language source to Chinese


tekezo:
Thanks!

PR #925, opened 2021-08-28
Updated complex modification: Capslock vim movements to rev2

paulchiu:
Changes:

- Split out "caps + enter" to trigger caps combo into own rule.
- Added "caps + a" to trigger caps combo rule.


tekezo:
Thanks!

PR #934, opened 2021-09-06
add new rule: emacs key for Obsidian.app

hnsol:
add rules: emacs like key binds only for Obsidian.app, at vim-insert mode


MuhammedZakir:
Merged. :-)

PR #935, opened 2021-09-07
Add new modifications rule: rdp-japanese-us

willow-micro:
This rule can make following modifications:

- Send Japanese Eisu/Kana when Left/Right Command was pressed
   - Same as _For Japanese （日本語環境向けの設定） (rev 6)_ in the International (Language Specific) category
   - But only if Microsoft Remote Desktop or Parallels Client is not the frontmost application
- Swap Command and Option Keys
   - If Microsoft Remote Desktop or Parallels Client is the frontmost application


tekezo:
Thanks!

PR #938, opened 2021-09-09
Move for_korean_keyboard rules from Modifier Keys to International (Language Specific)

Metalliza:
It should be categorized into `International(Language Specific)`.


tekezo:
You're right, thanks!

PR #939, opened 2021-09-13
change rules: emacs like key binds only for Obsidian.app, at vim-insert mode (rev02)

hnsol:
Deleted `"optional":  "any"` in modifiers,
to avoid working when users press anything other than the control key.

### DIFF
--- a/public/json/obsidian_emacs_vim-insert.json
+++ b/public/json/obsidian_emacs_vim-insert.json
@@ -1,5 +1,5 @@
 {
-  "title": "emacs like key for Obsidian.app_rev02",
+  "title": "Obsidian.app, emacs like key at vim-insert mode_rev02",
   "rules": [
     {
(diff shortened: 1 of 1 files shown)


public/json/obsidian_emacs_vim-insert.json
```diff
@@ -12,9 +12,7 @@
               "mandatory": [
                 "left_control"
               ],
-              "optional": [
-                "any"
-              ]
+              "optional": []
```
MuhammedZakir, line 15: Empty fields are not needed. You can remove them.

MuhammedZakir (changes requested):

public/json/obsidian_emacs_vim-insert.json
hnsol: Thank you for your review and comments. I fixed them.

MuhammedZakir:
Merged. :-)

PR #940, opened 2021-09-16
Add 'Colemak Mod-DHk'

xsc:
According to the [Colemak Mod-DH Revision History](https://colemakmods.github.io/mod-dh/#revision-history) the original layout for staggered keyboards - where M and K were swapped - is now known as `Colemak Mod-DHk`.

While no longer recommended for use, there might still be people - like me - that have gotten accustomed to it. This PR adds this alternative layout into the `Colemak Mod-DH` pack.

### DIFF
--- a/public/json/colemak_dh_ansi_layout.json
+++ b/public/json/colemak_dh_ansi_layout.json
@@ -10779,6 +10779,3599 @@
           ]
         }
       ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


MuhammedZakir:
Merged. Thanks! :-)

Note: I didn't check whether the layout is correct (according to spec). I'm trusting you on that. ;-)

PR #948, opened 2021-10-05
Updated product_id to match Tartarus Pro

2Wdavidcunliffe:
- This new profile is a copy of the[ Razor Tartarus V2](https://github.com/pqrs-org/KE-complex_modifications/blob/main/public/json/Razer_Tartarus_v2.json) profile but updated to match the Tartarus Pro series device


MuhammedZakir:
Is the configuration same? If so, you can just add another device identifier for Tartarus Pro to `device_if` conditions. No need for a new config if configuration is same.

Ref: https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/conditions/device/#multiple-entries

```json
{
  "type": "device_if",
  "identifiers": [
     {
       <identifiers for a device>
     },
     {
       <identifiers for another device>
     }
  ]
}
```

PR #962, opened 2021-11-04
Add the option to open a terminal window using Ctrl-RAlt-T.

sgielen:


public/json/open_terminal_window.json
```diff
@@ -0,0 +1,31 @@
...
+                "from": {
+                    "key_code": "t",
+                    "modifiers": {
+                        "mandatory": [
+                            "left_control",
+                            "right_option"
+                        ],
+                        "optional": [
+                            "any"
+                        ]
```
MuhammedZakir, line 17:
```suggestion
                        "optional": [
                            "caps_lock"
                        ]
```

Using `any` in `optional` modifiers will execute this manipulator for all `t+<modifiers>` combo. Either don't use any `optional` modifiers, or only add `caps_lock`, so this will work when caps lock is on.

sgielen: Perfect, sounds like an improvement indeed, thank you!

MuhammedZakir:
Thanks! :-)

PR #963, opened 2021-11-05
Add PC style word navigation rules

mbaker3:
Users who want to have PC style selection likely also want PC style navigation.

### DIFF
--- a/public/json/Richi_Select_Word_PC.json
+++ b/public/json/Richi_Select_Word_PC.json
@@ -1,5 +1,5 @@
 {
-  "title": "Richi - Select and Navigate Word PC Style (you must use swape COMMAND and CTRL in other case)",
+  "title": "Richi - Select and Navigate Word PC Style (you must use swap COMMAND and CTRL in other case)",
   "rules": [
     {
       "description": "richi - command+shift+right To options+shift+right ",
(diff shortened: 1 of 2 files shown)


MuhammedZakir:
I think we need to duplicate all manipulators, but without `shift` in modifiers, to get all navigations. What do you think?

mbaker3:
I'm not sure what the intention for the original up/down mappings was. I don't use those shortcuts and a quick test in Notepad on Windows revealed that `Ctrl+Shift+Up/Down` and `Ctrl+Up/Down` don't appear to do anything.
In Chrome the key combination is no different than pressing `Shift+Up/Down` or `Up/Down`

Do you know what the expected behaviour for `Ctrl+Shift+Up` and `Ctrl+Shift+Down` is in Windows?

I'm happy to setup mappings if I know what behaviour I'm trying to emulate!

MuhammedZakir:
I can't test them as I don't use Windows. But, even if they don't work in Windows, IMHO, if shortucts work in macOS, then we should just leave them as is. Although this may go against "_PC style_ word navigation".

MuhammedZakir:
So what should we do about this:

> I think we need to duplicate all manipulators, but without shift in modifiers, to get all navigations. What do you think?

If you are against this, I can merge your PR as is.

mbaker3:
Based on our discussion I don't think it makes sense to duplicate all manipulators without the shift modifiers.

**Reason:** Windows users don't expect any behaviour for that key combination.

mbaker3:
Just noticed the original title had a typo in it. Added the fix to this PR.

MuhammedZakir:
Thanks! :-)

PR #975, opened 2021-11-11
Add the option to mute teams, zoom or slack using capslock key

aceslick911:


MuhammedZakir:
1. Please add your config to `public/groups.json` in "Application-specific" category.
2. `"modifiers": []` is not needed. You can remove them.
3. `repeat` is by default `true`. So no need for that.

4. why do you use `lazy` flag in those rules?

Moulick:
Should instead https://github.com/pqrs-org/KE-complex_modifications/blob/main/public/json/f5_ms_teams_zoom_mic_mute.json be updated to support slack rather than a new one?

PR #978, opened 2021-11-17
Fn keys working without having to press fn in JetBrains Rider

ArtS:
As per title -- I got sick of having to press fn every time when using F keys in OS X


MuhammedZakir:
Thanks!

PR #983, opened 2021-11-26
Adds shortcut to open Alacritty terminal

dnmellen:


MuhammedZakir:
IMHO, it is better to add this is as another rule in `open_terminal_window.json`. What do you think?

dnmellen:
I tried to do so, but it fails when I try to run `make` again. The flow of adding a new rule is not very clear in the documentation. I followed the docs the first time and I created a `open_alacritty_terminal.json.rb` under `src/json` the first time, then I ran make and it built ok, but now I can't see the `.rb` file and if I just remove `public/json/open_alacritty_terminal.json` and modify `open_terminal_window.json` then the make fails because it still requires me to add `open_alacritty_terminal.json` in `groups.json`

MuhammedZakir:
It shouldn't. See the GHA ran for a testing PR I made. All the tests passed successfully. I think you may have made some mistakes when you tried to do this.

GHA: https://github.com/MuhammedZakir/KE-complex_modifications/runs/4341823190?check_suite_focus=true
PR: https://github.com/MuhammedZakir/KE-complex_modifications/pull/1

PR #991, opened 2021-12-02
Add Finder improved Navigation rule

honsa:
All the available definitions do not work onedit mode. 

This modification will work around that uses a variable on rename F2 to indicate we are on edit.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -505,6 +505,9 @@
         },
         {
           "path": "json/jetbrains_rider_fn.json"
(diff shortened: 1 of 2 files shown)


honsa:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#approving-workflow-runs-on-a-pull-request-from-a-public-fork

tekezo:
Thanks!

PR #992, opened 2021-12-03
A rule to reconfigure Russian keyboard from Mac layout to Windows layout

ov7a:
Mac layout for Russian language looks like this:
image

It is a pain for ex-non-mac-user to type symbols like . or , in unusual place. Also, as far as I know, Mac doesn't support different layouts for different inputs, so it would be even more frustrating to use external non-mac keyboard, like this one:
image

This PR is an attempt to make external keyboard more pleasant to use in Mac. There are two rules, each remaps punctuation and some special symbols to alternative ("windows") key combination. The only difference is that one rule works for non-Mac keyboards ("I want external keyboard to behave like I used to, but Ma Laptop keyboard should work as intended"), while another works for all keyboards ("I touch type well and don't want to learn new layout").

Unfortunately I hasn't found a way to emit backslash in Russian layout, and remapped it to keypad_slash.

Tested with MacBook Pro Retina 15" A1398 and Dell keyboard KB212-B.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -696,6 +696,9 @@
         {
           "path": "json/rdp-japanese-us.json",
           "extra_description_path": "extra_descriptions/rdp-japanese-us.json.html"
+        },
+        {
+          "path": "json/russian-keyboard-remapping.json"
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #996, opened 2021-12-05
Adds RightOptionFN

jcarlos:
Makes common programming symbols easily accessible on the home row of the Lenovo Trackpoint II keyboard, or any other keyboard where the right alt is easily reachable with the right thumb.


tekezo:
Thanks!

PR #997, opened 2021-12-06
Add KanaFN

kotarok:
KanaFN is similar to SpaceFN but more comfortable for Japanese keyboard users, because Kana key is placed more thumb friendly position on Japanese keyboard.

Provides those settings below:
- KanaFN: japanese_kana to fn unlesss alone
- KanaFN: Kana+[ijkl] to Up, Left, Down, Right
- KanaFN: Kana+[h;] to Backspace, Delete
- KanaFN: Kana+[op] to Page Up, Page Down
- KanaFN: Kana+[,.] to Home, End
- KanaFN: Kana+N to Enter
- Optional: spacebar to fn unlesss alone
- Optional: japanese_eisuu to left_option unlesss alone

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -712,6 +712,9 @@
         },
         {
           "path": "json/russian-keyboard-remapping.json"
+        },
+        {
+          "path": "json/kanafn.json"
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #998, opened 2021-12-07
Create fn+up-to-down.json

tonysleva:
Made a different version of ctrl+up json so i could use the fn key instead.


tekezo:
Thanks!

PR #999, opened 2021-12-07
Add VSCode exception to auto_close_brackets_and_quotes

hzpc-joostk:
Adds Microsoft's editor VSCode `com.microsoft.VSCode` to the conditional exceptions of `public/json/auto_close_brackets_and_quotes.json`, so closing brackets and quotes are not automagically added in that editor.


tekezo:
Thanks!

PR #1002, opened 2021-12-09
Add complex modification to disable WiFiDiagnostics shortcut

qubist:
See this AskDifferent thread for context: https://apple.stackexchange.com/questions/373378


tekezo:
Thanks!

PR #1004, opened 2021-12-13
Add Complex Modification to Open control center with f8

kerry-zhu:
Maps the f8 button to fn+c, which is the Mac keyboard shortcut for opening the control center


tekezo:
Thanks!

PR #1007, opened 2021-12-15
Add "section_sign_to_forward_delete" modification

Radllaufer:
It expands on the simple modification of "non_us_backslash/section sign to delete" with deleting up to the next space and the whole line using "option + §" and "command + §" respectively.


tekezo:
Thanks!

PR #1008, opened 2021-12-15
Add Swap "command + tab" with "control + tab" mod

Radllaufer:
This will make it easier to switch between tabs inside applications.


tekezo:
Thanks!

PR #1018, opened 2021-12-26
fix(rules): Remove comment from groups JSON file

paulchiu:
Comments are not valid JSON. Removing what appears to be a copy and paste error.


tekezo:
Thanks!

PR #1019, opened 2021-12-26
Add "Fn/Caps + WASD(↑←↓→) / QE(Home End) / RF(PageUp PageDown)" Rule

hyacz:
This rule is inspired by "Caps Lock W/A/S/D Arrows" and replaced the extra layer with the form of combination keys. Replace Caps with Fn, then map Fn + W/A/S/D/Q/E/R/F to ↑/←/↓/→/Home/End/PageUp/PageDown.

Contains 4 rules:
- Change Caps to Fn
- Change Fn + W/A/S/D to ↑/←/↓/→
- Change Fn + Q/E to Home/End
- Change Fn + R/F to PageUp/PageDown


tekezo:
Thanks!

PR #1035, opened 2022-01-24
Add a Change Grave Accent(`) to Esc (Filco Minila-R) mapping.

RishyPa:
Grave Accent to Esc
option + Grave Accent to Grave Accent
~ still shift + Grave Accent


tekezo:
Thanks!

PR #1037, opened 2022-01-26
Space-cadet shift

akhar:
Left shift, if pressed only, for open parenthesis; right shift, if pressed only, for close parenthesis.


tekezo:
Thanks!

PR #1039, opened 2022-01-30
Add exclusions for parsec to windows_shortcuts_on_macos.json

ChristopherHaws:
Add exclusions for [Parsec](https://parsec.app/) since it essentially the same thing as RDP. The app is a high performance remote desktop app used for connecting to other computers. The most common use case is to connect to a Windows computer to play games, but can also be used for normal remote desktop activities.


akhar:
Thanks!

PR #1045, opened 2022-02-09
Keychron K2 rules to improve US keyboard layout usage for RU users

ikenfin:
Modification adds two simple key remaps to keep important shortcuts as close as possible between keyboards (US <-> RU)

1. Use ``` cmd+` ``` as `cmd+§`, because they are on similar place at keyboards
2. Use `option+tab` as ``` cmd+` ```, because there is no similar key on US keyboard (_key next to left shift_), so use new shortcut


akhar:
Thanks!

PR #1046, opened 2022-02-10
Escape toggles capslock if enabled

aalvarado:
If Capslocks is on, and if you press escape, you will disable caps


akhar:
You forgot file extension. Please, rename `public/json/esc-caps-toggle` to `public/json/esc-caps-toggle.json`

tekezo:
Thanks!

PR #1063, opened 2022-03-03
Add rule: Use F12 to show/hide iTerm (Quake-style "instant terminal")

n-st:
Perhaps a very specific use case, but the general launch/hide functionality might come in handy for others.

The F12 key can be a bit problematic, but is personal muscle memory from my time with [YuaKuake and Guake](http://freesoftwaremagazine.com/articles/how_to_use_quake-style_terminals_on_GNU_Linux/).

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -549,6 +549,9 @@
         },
         {
           "path": "json/ms_teams_ptt.json"
+        },
+        {
+          "path": "json/f12-show-hide-iterm.json"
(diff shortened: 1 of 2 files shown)


n-st:
… and I just noticed that iTerm has an integrated "Hotkey Window" feature that does something similar to this. Might be useful for some users as well.

tekezo:
Thanks!

PR #1068, opened 2022-03-06
PC-style delete

vladfi1:
The PC-style shortcuts currently only work for backspace, but not delete. This PR enables control+delete to work too.


tekezo:
Thanks!

PR #1069, opened 2022-03-08
Fix Neo2 layer 4 rules

### LINKED ISSUE #1057: Neo2 - layer 4 and layer 6 not working in latest version
I had these modifications installed before:

- Neo2 mod 3 and 4 keys (Apple keyboard)
- Neo2 layer 4
- Neo2 layer 6
- Toggle caps_lock by pressing left_shift + right_shift at the same time

I just tried to update to the latest versions available on ke-complex-modifications.pqrs.org, but after the upgrade, layer 4 and layer 6 are not working properly any more.

Using the left mod4 key does not work on all keys. For example, mod4+j (n on Neo2) correctly results in `4`, but mod4+h (s on Neo2) gives me `S` instead of `¿`.

Pressing the left mod3 and mod4 key together should get me into layer 6, however, it seems that this brings into layer 5 instead.

Reverting `Neo2 layer 4` and `Neo2 layer 6` to the previously installed versions fixes both problems. I'm using the internal MacBook Pro keyboard with these modifications.

In the same vein, the previous `Neo2 mod 3 and 4 keys (Windows keyboard)` modification worked fine in combination with a Keychron K6 keyboard, but the newer `Neo2 mod 3 and 4 keys (Windows keyboard). Rule applied to all non-Apple keyboards` modification does not bring me to layer 3 with the left mod 3 key.

jgosmann:
Partially reverts 436d4f944a2a58b8295244a20f52a488a6c5c596 which changed
the `to_pre_events` keys for unclear reasons.

Fixes #1057.

# Todo

* [ ] Update the "extra description" as the rule names in there are not update and the list is incomplete.

### DIFF
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -1515,7 +1515,7 @@
           },
           "to": [
             {
-              "key_code": "keypad_1",
+              "key_code": "z",
               "modifiers": [
                 "left_option",
                 "left_shift"
@@ -1543,7 +1543,7 @@
           },
           "to": [
             {
-              "key_code": "keypad_1",
+              "key_code": "z",
               "modifiers": [
                 "left_option",
                 "left_shift"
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #1072, opened 2022-03-11
Tap CapsLock send Escape, Held CapsLock send Ctrl

dllvhaobo:
Send Escape if only CapsLock tapped.
Send Left_control if combined with other key pressed.


tekezo:
Thanks!

PR #1073, opened 2022-03-11
create dead key map for open bracket; quote; hyphen

delow:
This mapping of keys close bracket, equal, and semicolon with command and command+shift will work around 3 dead keys on my apple keyboard.


tekezo:
Thanks!

PR #1075, opened 2022-03-12
add caps_lock_magic.json Semicolon Pull Down Symbol

miozus:
* map holding semicolon to a get symbol from shift 1~9, etc , it is very convenient for vimer , especially who use the 60 keys keyboard.

example:

hold on `;` and press `a`  then output `*`

---

ps.All my effort is the parody of Vonng(caps_lock_enchancement)  & Salted Fish Akang.

I like capslock Enchancement , so I extend it, as part of it, called Magic.

I also extend it with AutoHotkey on Windows 11.  See [miozus/CapslockMagic](https://github.com/miozus/CapslockMagic)


tekezo:
Thanks!

PR #1077, opened 2022-03-13
update caps_lock_magic.json (fix): 3 awake digital keyboard

miozus:
* fix 3 + ; = enter
  Cause by typo in escape character in yaml sytax quote -> double quote

* fix details formmater at html docs


miozus:
Thanks!

PR #1081, opened 2022-03-19
Added a shift-delete to forward delete shortcut

orlp:
I am aware that there's a couple of these already, but none had the behavior I wanted, especially with regards to Command-Shift-Delete becoming Ctrl-K to delete until the end of the line.


tekezo:
Thanks!

PR #1084, opened 2022-03-25
Add switchable vimish navigation rules

jiyolla:
3 different rules to toggle vimish navigation mode
currently only 1 rule to apply under vimish navigation mode which is to
hjkl to arrows.

I also map capslock to control when using this mode.


tekezo:
Thanks!

PR #1085, opened 2022-03-25
Added a json allowing the use of a single keystroke for tildes and backticks with the French PC keyboard layout

Litarvan:
Using a keyboard layout from a language with accents (French, Spanish, ...), when accents keys (tildes, backticks, etc.) are pressed macOS will wait for the next letter to maybe 'merge' them (ex: pressing backtick then `e` makes `è`). If it cannot merge them, it will sometimes have a strange behavior, like doubling the tilde/backtick, which is very annoying.

I added a JSON that prevents macOS from waiting a key after typing a backtick or a tilde using the French PC layout, by pressing the right arrow right after.


tekezo:
Thanks!

PR #1086, opened 2022-03-26
Shapr3d middle mouse button rotate

jrsall92:
In Shapr3d the middle and right mouse buttons do the same(pan). I mapped the middle button to l_shift + right button (roation). I suggest dropping the held down threshold millis to 100 for snapier activation


tekezo:
@jrsall92 
Thank you for PR!

The renamed json file is not included in commit 7d9f3a0.
Could you add the json file?

jrsall92:
Sorry @tekezo, it's added now, thanks for your time

tekezo:
Thanks!

PR #1087, opened 2022-03-27
add hhkb_fn_comma_period_to_ltgt.json

shinouta:
I have added a json file "hhkb_fn_comma_period_to_ltgt.json".
This file is only for HHKB devices.
It converts Fn+comma(End) and Fn+period(PgDn) to less than(<) and greater than(>) respectively.


tekezo:
Thanks!

PR #1089, opened 2022-03-29
Add multiple_input_sources rule

blwinters:
Adds rule to allow toggling between multiple input sources with right_option and a number key.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -683,6 +683,9 @@
           "path": "json/czech_pc_shortcuts.json"
         },
         {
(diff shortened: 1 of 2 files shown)


blwinters:
I realized that this rule has a bug, at least for switching between two Chinese input sources. It seems that the text field needs to lose focus in between switching keyboards in order for the new input source to be active. Clicking the menu bar to switch sources or using system short cuts to toggle from the overlay menu achieves this loss of focus, my current pull request does not.

PR #1091, opened 2022-03-29
Create swap-command-tab-with-control-tab-optional-shift.json

7fe:
I have control and command swapped around on my Macbook Pro.

This and iTerm(which swaps command and control back) fixes my major problem.

This is meant for toggling between windows.


tekezo:
Thanks!

PR #1092, opened 2022-03-30
Create ScreenshotToBackspace.json

AustinTSchaffer:
Remaps command+shift+5 to backspace. Just got a Keychron K3 and I keep hitting the screenshot key when I'm trying to hit backspace.


tekezo:
Thanks!

PR #1095, opened 2022-04-03
Adding RCMD to LCMD+OPT

Rguarnizo:
right cmd to left cmd + option key, util for swap tabs in browers, use with HJKL or WASD for improve functions.


tekezo:
Thanks!

PR #1097, opened 2022-04-08
Add personal rules for iwaltgen

iwaltgen:
File with personal rules.
- Vi Mode
- Forward delete
- Right command

really good to be able to store personal rules 🙂

thanks!


tekezo:
Thanks!

PR #1102, opened 2022-04-17
Support javascript generator

tekezo:


yeshiqing:
【Suggestion】
It should be mentioned in the part `Add rules` of `README.md` that the JSON generator file’s filename must be the `*.json.*`, otherwise the JSON file can't be generated.

PR #1103, opened 2022-04-17
Add comfortable_ctrl_w_to_delete_backward_word.json

andmis:
Hi,

I'm writing a couple of blog posts about keyboard remapping and it would be really great if I could point readers to a complex modification that implements what my post suggests. Here's a draft if you're interested: https://andmis.com/km/ctrl-w/

Thanks


tekezo:
Thanks!

PR #1105, opened 2022-04-19
Add German Characters (Umlauts and Eszett) with Caps Lock

justinmcbride:
Add a new Complex Modification that hosts a set of different rules.
Use the Caps Lock key as a modifier key, but for getting access to the German letters. Easy access to äöü, ÄÖÜ, and ß.

### DIFF
--- a/public/extra_descriptions/german_umlauts_caps.html
+++ b/public/extra_descriptions/german_umlauts_caps.html
@@ -11,6 +11,7 @@
 <h5>What Keyboards are Supported</h5>
 <p>Any keyboard with the standard Latin characterset is supported. The purpose of this modification is to enable use of the German characters on a keyboard that doesn't have them.</p>
 <p>As long as you have a Caps Lock key, as well as [aous], you're good to go.</p>
+<p>That said, it's limited to ANSI or ISO keyboards with an English input language.</p>
 
 <h5>Motivation</h5>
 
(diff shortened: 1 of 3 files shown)


MuhammedZakir:
It maybe better to specify which type of keyboard layout is needed for this to work.

justinmcbride:
> It maybe better to specify which type of keyboard layout is needed for this to work.

Thanks for the suggestion. Can you provide guidance on how you propose I do this? Simply inside the extended info?

MuhammedZakir:
1. Yes, explain that in the extended info.
2. Optional, but recommended: if you can restrict the manipulators to certain layouts (e.g. German language?) [1] or certain keyboard types (e.g. ISO) [2], then do so. Don't need to language restriction if you want to and can use those in other keyboard layouts.
    - The reason for this is to prevent conflicts when you have or imported many rules.

[1] https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/conditions/input-source/
[2] https://karabiner-elements.pqrs.org/docs/json/complex-modifications-manipulator-definition/conditions/keyboard-type/

justinmcbride:
> 1. Yes, explain that in the extended info.
Went ahead and added that info.

> 2. Optional, but recommended: if you can restrict the manipulators to certain layouts (e.g. German language?) [1] or certain keyboard types (e.g. ISO) [2], then do so. Don't need to language restriction if you want to and can use those in other keyboard layouts.

I added those restrictions.

MuhammedZakir:
Merged. :-)

PR #1106, opened 2022-04-19
Only show the extended information once per Complex Modification file

justinmcbride:
Modify the CategoryBox to only show the extended information once per Complex Modification file.

Because there can only be a single extended information file for each submitted JSON file, it doesn't make much sense to repeat that `iframe` for every rule within that single, submitted file. 

Here is a screenshot of it before:
image

And after my changes:
image


tekezo:
Thank you for your PR!
I merged this PR with a small change that fixed `available_since` position.

Screen Shot 2022-04-20 at 13 03 52
Screen Shot 2022-04-20 at 13 04 22

PR #1115, opened 2022-05-13
Update fn+cmd_to_eisu_and_kana.json

BaksiLi:
A keymap json to map `fn + left_/right_command to japanese_eisuu/kana`. 

This helps to mimic the JIS layout's feeling on other keyboard layouts.


tekezo:
Thanks!

PR #1117, opened 2022-05-13
Add keymap to prevent exiting fullscreen on escape

awesmubarak:
This keymap prevents exiting full screen on escape while (seemingly) retaining other escape functions. I have tested escape function retention with full screen videos in safari and firefox, text input boxes, and vim in iTerm 2. The prevention of full screen exit seems to work across all apps that supported it.

This is similar to #109, but with two differences:

- It works across all apps
- It uses `shift-escape` instead of `option-escape`. Using `option` didn't prevent exiting for me on macOS Catalina.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -846,6 +846,9 @@
           "path": "json/esc_to_eng_ime.json"
         },
         {
+          "path": "json/esc_to_shift_esc.json"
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #1122, opened 2022-05-19
add shift key training wheels 2, habit breaking while still intelligible 

Willardgmoore:
add shift key training wheels 2, more forgiving version of training wheels where letter is still typed though it is lowercase. Great for typing quick and still understanding what you intended to write while breaking bad habits.


tekezo:
Thanks!

PR #1130, opened 2022-05-24
Add ADNW to supported Neo2 layouts

elKei24:
[ADNW](http://adnw.de) uses the mod layers of Neo. Therefore, it was recommended to use the Neo2 karabiner rules for ADNW. This is not working any more because the rules are only applied with Neo keyboard layouts. This PR adds the ADNW layout as a compatible layout that can be used with the rules.

### DIFF
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -34,6 +34,9 @@
                 },
                 {
                   "input_source_id": "^org\\.unknown\\.keylayout\\.DeutschNeoQwertz$"
+                },
+                {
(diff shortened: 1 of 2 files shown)


MuhammedZakir:
Can you update your branch by rebasing? You can do that using `Update branch` option beside the branch info below.

elKei24:
> Can you update your branch by rebasing? You can do that using `Update branch` option beside the branch info below. I can do that if you're fine with it.

I did, thanks for pointing that out. Feel free to modify this branch yourself in any way.

PR #1132, opened 2022-05-29
Keychron k6 add device conditions

sourabhv:


public/json/keychron-k6-shortcuts.json
```diff
@@ -1,17 +1,29 @@
...
         {
+          "conditions": [
+            {
+              "type": "device_if",
+              "identifiers": [
+                {
+                  "product_id": 591,
+                  "vendor_id": 1452,
+                  "is_pointing_device": false,
+                  "is_keyboard": true
```
MuhammedZakir, line 17: Either one is enough.

public/json/keychron-k6-shortcuts.json
```diff
@@ -1,17 +1,29 @@
...
+            }
+          ],
           "type": "basic",
           "from": {
             "key_code": "escape",
             "modifiers": {
-              "mandatory": [
-                "shift"
-              ]
+              "mandatory": ["shift"]
```
MuhammedZakir, line 26: Why is this changed? This is not needed as it will be changed when this file is formatted.

sourabhv: My formatter changed this, its fine if it will be reverted

PR #1138, opened 2022-06-03
Windows-like key combinations for French-AZERTY layout

Louis-DR:
left_command 4  =>  {
left_command -_  =>  }
left_command 5  =>  [
left_command )°  =>  ]
left_command 6  =>  |
left_command 8  =>  \\
§  =>  -


MuhammedZakir:
Isn't it much better to merge all these into a single file? You can and (imo) should use multiple rules for config like this.

Louis-DR:
Ready to be merged

PR #1144, opened 2022-06-18
Using right option as Hyper key

tockrock:
Simple rule to remap right option to hyper key

Based on ctrl_to_hyper.json by @luanzeba


tekezo:
Thanks!

PR #1146, opened 2022-06-23
add rshift home/end to rshift + right cmd to select to start/end line

holmestm:
Existing shift + home/end doesn't modify right shift key on my extended external apple keyboard


tekezo:
Thanks!

PR #1147, opened 2022-06-23
Caps lock to backspace and backspace to delete forward, but only on Colemak

DanielZTing:
This is useful to me because sometimes I need to let other people use my computer and I no longer have to quit Karabiner to undo Colemak's caps lock to backspace mapping.

### DIFF
--- a/public/json/colemak-caps-lock-backspace-delete-swap.json
+++ b/public/json/colemak-caps-lock-backspace-delete-swap.json
@@ -1,0 +1,69 @@
+{
+  "title": "Caps lock to backspace and backspace to delete forward, but only on Colemak",
+  "rules": [
+    {
+      "description": "Caps lock to backspace, but only on Colemak",
+      "manipulators": [
+        {
+          "type": "basic",
+          "from": {
+            "key_code": "caps_lock",
+            "modifiers": {
+              "optional": [
+                "any"
+              ]
+            }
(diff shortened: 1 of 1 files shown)


public/json/colemak-caps-lock-backspace-delete-swap.json
```diff
@@ -0,0 +1,73 @@
...
+                "any"
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "delete_or_backspace",
+              "lazy": false,
+              "repeat": true
+            }
```
MuhammedZakir, line 22: You don't need to add these as these are default values. Please remove them.

public/json/colemak-caps-lock-backspace-delete-swap.json
```diff
@@ -0,0 +1,73 @@
...
+              "optional": [
+                "any"
+              ]
+            }
+          },
+          "to": [
+            {
+              "key_code": "delete_forward",
+              "lazy": false,
+              "repeat": true
```
MuhammedZakir, line 55: This too.

MuhammedZakir (changes requested):

MuhammedZakir (approved):

MuhammedZakir:
Merged. :-)

PR #1149, opened 2022-06-24
Add new script to use caps_lock as fn when pressed with other keys

yuntaechriskim:


yuntaechriskim:
Thought I found some unexpected behavior, but seems to work fine as is

tekezo:
Thanks!

PR #1155, opened 2022-07-05
Add rule: Section sign (§) to backtick (`) in English keyboards 

kuzdogan:
Based on @fredericrous's [setting](https://github.com/pqrs-org/KE-complex_modifications/blob/main/public/json/section_sign_to_backtick.json) but with English keyboard condition. Useful for people using multiple layouts.


tekezo:
Thanks!

PR #1156, opened 2022-07-08
Add Option+ PageUp, PageDown, Home, End to Control+Arrows

markomarkovic:


MuhammedZakir:
Why `any` in `from.modifiers`? `caps` is okay, but adding `any` will also modify `<other modifiers>+option+<key>`, which doesn't seem to be what you want?

MuhammedZakir:
Merged. :-)

P.S. Sorry about that "caps". :-)

PR #1172, opened 2022-08-01
Add azerty_style_windows_keyboard_on_mac.json

NotCassim:
A small new json very useful for french users on Mac. They will be able to easily use french quote « and » with a quick shortcut thanks to KE. Additionally, it also provides rules to create brackets faster [].


tekezo:
Thanks!

PR #1173, opened 2022-08-01
Add focus desktops with option + numbers

dmun:
Enables you to use option + numbers to switch desktops instead of control + numbers. When using Yabai without SIP disabled you normally wouldn't be able to manually focus a desktop. Option + [key] is also a common bind combination for Yabai users.


tekezo:
Thanks!

PR #1174, opened 2022-08-02
add UTM to VM list

russell:
UTM is a wrapper around QEMU https://mac.getutm.app it provides a GUI
experience using Spice.


tekezo:
Thanks!

PR #1177, opened 2022-08-05
Add a modernized version of Apple Music.app-only media key control

ambv:
There's already a `public/json/media_key_only_control_itunes.json` modifier but it's got two problems:
- the application to control is no longer called "iTunes" so it doesn't work
- it refers to keys as F7 - F9 keys which is meaningless for Touch Bar controls

This PR introduces a new modifier that works with Apple Music and provides refined docs.

### DIFF
--- a/public/extra_descriptions/media_key_only_control_itunes.html
+++ b/public/extra_descriptions/media_key_only_control_itunes.html
@@ -4,10 +4,15 @@
   Description
 </p>
 <p>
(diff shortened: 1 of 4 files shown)


ambv:
This is how it looks:
Screen Shot 2022-08-05 at 21 49 21

tekezo:
Thanks!

PR #1190, opened 2022-08-13
feat: add language switch with control + enter

ptrln216:
Add support to use control + enter to switch language. Helpful for users who has the need to switch input languages often.


tekezo:
Thanks!

PR #1194, opened 2022-08-22
Use Right Shift + numbers as Function keys

witalewski:
On 60%-style keyboards, you usually use a function key + number keys to emulate F1, F2, ..., F12 keys.
On macOS specifically this will normally mean that you can use this combination for *either* actual function keys *or* media keys.
Assuming that the System Preferences are set to use F-keys as media keys, this rule allows you to use right shift + number keys as regular F1 - F12 function keys.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -173,6 +173,9 @@
         },
         {
           "path": "json/meh_capslock.json"
+        },
+        {
+          "path": "json/right_shift_plus_numbers_to_function_keys.json"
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!
I moved this setting into the `Key Specific` group.
https://github.com/pqrs-org/KE-complex_modifications/commit/2637b71521fa3447aa247d05a97e2cdebd911148

PR #1198, opened 2022-08-26
pc printscreen for 3-way capture

ziippy:
I want using capture with 3-way mode.

so, I made the json for below

printscreen -> command+shift+3
alt+printscreen -> command+shift+4
ctrl+printscreen -> command+shift+5


tekezo:
Thanks!

PR #1206, opened 2022-09-06
Add custom TES68 volume brightness mods

willacai:
These rules convert the original Page Up/Down keys on a TES68 to be Volume and Brightness mod keys, respectively. Hold down Page Up/VMod and use up/down arrow keys to control volume, and the same for Page Down/BMod.


tekezo:
Thanks!

PR #1208, opened 2022-09-08
Double tap keypad_enter to double tap fn (start dictation.) Single tap for keypad_enter after a short delay (250ms.)

ndbroadbent:
This rule solves a problem that I've been having with my Microsoft Surface Ergonomic Keyboard ([see my question on Apple StackExchange](https://apple.stackexchange.com/questions/444325/how-can-i-configure-a-key-to-start-dictation-with-a-microsoft-surface-ergonomi).)

The `Fn` key can't be used to start dictation, since it only toggles an internal state on and off, and it does not actually send any keypress event. This rule allows you to start dictation by double tapping the keypad enter key, which will send Fn twice. If you only press the keypad enter key once, it will send the original enter key after a short delay (250ms). Note that the first keypress event will always send Fn, but this usually won't do anything by itself.

I chose the keypad enter key because it is in a very convenient position at the bottom right of the keyboard. I also use this key is when I'm entering the PIN for my Yubikey, so I still wanted to be able to press enter occasionally.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -65,6 +65,9 @@
           "path": "json/double_tap_ctrl_to_cmd.json"
         },
         {
+          "path": "json/double_tap_keypad_enter_to_double_tap_fn.json"
+        },
+        {
           "path": "json/f4_to_cmd_space.json"
         },
         {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #1213, opened 2022-09-12
Add Durgod Hades UK Layout

sammckay10:
Hi, 

I've added some mappings for the Durgod Hades UK layout. All Keys now match their printed letters.


tekezo:
Thanks!

PR #1217, opened 2022-09-21
Add Rule of Ctrl + g = Escape

noblejasper:
Added `keyboad-quit`, which is Emacs' way of performing escaping with Ctrl + g.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -176,6 +176,9 @@
         },
         {
           "path": "json/meh_capslock.json"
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

I moved this rule into Key Specific group by this commit.
https://github.com/pqrs-org/KE-complex_modifications/pull/1218/commits/1f16e86817882e39eaa90c60320ad0cba1081366

PR #1224, opened 2022-10-04
feat: Add complex modification - map right option key to underscore

jtoaha:
Hello. Added a complex modification - mapping the right option key to underscore `_`. Motivation is more efficiency when typing, such as editing things like filenames and variable names when programming, etc...


tekezo:
Thanks!

PR #1226, opened 2022-10-11
fn + return to insert modifier

kaspervdl:
new modifier to map Fn + Enter to Insert


tekezo:
Thanks!

kaspervdl:
Thank you too!

PR #1228, opened 2022-10-13
add Personal rules (@toshi-ue)

toshi-ue:
add keyboard rules for me.
- switch numbers and symbols
- (JIS) in terminal app, shortcut command for save or discard changes
- thumb shift
- (JIS) Swap ¥ and Alt+¥


tekezo:
Thanks!

PR #1234, opened 2022-10-31
Add Zweih's Arrow Keys modification.

Zweih:
Adding layout for FN + IJKL to arrow keys and using FN+ Q to escape. This also accounts for other modifier keys being pressed (option, shift, GUI, etc.) to allow for arrow modifier actions.


tekezo:
Thanks!

PR #1235, opened 2022-10-31
Swap FN and left_command except in Terminals

Zweih:
Similar to [this modification](https://github.com/pqrs-org/KE-complex_modifications/blob/main/public/json/swap_ctrl_and_cmd_except_in_terminal.json), except it swaps with FN instead and it includes iTerm from [this PR](https://github.com/pqrs-org/KE-complex_modifications/pull/1079).


tekezo:
Thanks!

PR #1236, opened 2022-11-01
Added new emacs keybindings (port of universal emacs keybindings)

justintanner:
Hi There!

I ported my Hammerspoon emacs script to Karabiner, because Hammerspoon is now unusable on Ventura and Karabiner works great.

I created a new keybinding instead of adding to the existing emacs keybindings because I wanted to implement **set mark** outside of emacs and wanted to structure things under a single rule.

Let me know if I'm on the right track.


tekezo:
Thanks!

PR #1237, opened 2022-11-02
Added personal rules (@chxmvn)

CHMarvin:
Personal rules which support Ctrl + Y and Ctrl + Z from Windows and work with the German keyboard layout


tekezo:
Thanks!

PR #1239, opened 2022-11-07
Added redundant function to map caps lock to "apple_vendor_keyboard_key_code":"spotlight"

JackSalzman:


public/groups.json
```diff
@@ -1074,6 +1074,9 @@
         {
           "path": "json/CapsLockToSearch.json"
         },
+        {
+          "path": "CapsLockToSpotlightKey.json"
```
tekezo, line 1078: It should be `"json/CapsLockToSpotlightKey.json"`.

PR #1240, opened 2022-11-07
update personal rule for toshi-ue

toshi-ue:


toshi-ue:
I forgot to run the build.
So I tried `make package` in the project root to build, but it said `make: *** No rule to make target `package'. Stop.` and failed.
Why am I getting the error?

tekezo:
Thanks!

For `make`, just type `make` with no arguments.

PR #1246, opened 2022-11-11
Add Rule: Emacs key bindings (Move, Delete, Select)

devinmugen:
- Emacs key bindings (Move, Delete, Select)
	- Delete the previous word
	- Delete the next word
	- Move to the previous word
	- Move to the next word
	- Select the next word
	- Select the previous word
	- Select the current word


tekezo:
Thanks!

PR #1250, opened 2022-11-15
Add a full German ISO/PC mapping for PC keyboards on MacOS X

daemonkeeper:
Full ISO/PC mapping of "Windows" keyboards to MacOS X with a German layout


tekezo:
Thanks!

PR #1256, opened 2022-11-22
Logitech MX keyboard (connected via USB dongle). 

damianjankowski:
Add modification for Logitech MX keyboard (connected via USB dongle):

1. Swap left cmd with left option 
2. Swap right cmd with right option keys


tekezo:
Thanks!

PR #1258, opened 2022-11-24
Adding delete line mapping

iloveitaly:
adapted from https://stackoverflow.com/questions/47847591/karabiner-elements-delete-a-line


tekezo:
Thanks!

PR #1259, opened 2022-11-26
Add separated rules for Remote Desktop

beards:
Add some rules to keep usage habits as usual while using Remote Desktop.


tekezo:
Thanks!

PR #1272, opened 2022-12-11
fix tenkey numbers to use the keypad_[0-9] keys

ahornby:
This fixes the mapping for things like games and emulators that specifically want the keypad number keys


tekezo:
Thanks!

PR #1275, opened 2022-12-15
Type remote US keyboard via local JIS keyboard.

tonextone:
I do remote-desktop like this:

```mermaid
graph BT;
    A(local MacBook w/ JIS-keyboard) -. remote desktop .-> B(remote MacBook w/ US-keyboard);
```

When I'm doing remote-desktop with AnyDesk or "Google Remote Desktop",
my local JIS-keyboard behaves as if US-keyboard.

That was too annoying. So I made this.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -689,6 +689,9 @@
           "path": "json/jis_to_us_symbols.json"
         },
         {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #1281, opened 2022-12-21
feat: add mapping rules for tilde and grave accent

zhuo-cheng:
User Case:
HHKB Japanese layout with ANSI setup makes it impossible to enter tilde (~) and grave accent (`)

Solution:
* Map `Left Shift` + `Esc` to ~
* Map `Right Shift` + `Esc` to `


tekezo:
Thanks!

PR #1290, opened 2022-12-29
add: capslock mode

Titancube:
# Capslock Mode
Basic capslock enhancement mode for 65% ANSI layout keyboards with escape key containing tilde and swung dash

## Contents

- Caps lock to Hyper (Mandatory)
- hjkl to arrow keys
- Numbers to function rows (Useful if you have Nuphy 65% keyboards with function keys set to media keys)
- Tilde and Swung dash


tekezo:
Thanks!

PR #1291, opened 2022-12-29
add: mapped bindings for 12 keys on razer naga pro

mark-nicepants:
Default keys are a simple 1 to =, this maps them to add left_shift+left_option.

- mapped 9 to forward in chrome
- mapped 12 to back in chrome


tekezo:
Thanks!

PR #1295, opened 2023-01-08
Vim Mode Plus with Better Notifications

gilitos92:
A fork of the "Vim Mode Plus" modification by @harmtemolder that only modifies the way notifications work.


tekezo:
Thanks!

harmtemolder:
@gilitos92 If you want you can take over my "Vim Mode Plus" mod completely. I don't use macOS or Karabiner-Elements any more, so am not maintaining it...

gilitos92:
@harmtemolder Thank you for the trust and making the original script. I think it is better to leave them as separate Modifications because some people might prefer having the non-persistent notifications.

Din0Git:
Thanks for the script. Is there a way to change the placement for the notification? An offset of some sort? 

Greetz Dino

PR #1297, opened 2023-01-10
Added RDP rules for British/UK keyboard layout

bameyrick:
Specific mapping for mapping MacOS with British/UK keyboard layout to Microsoft RDP


tekezo:
Thanks!

PR #1304, opened 2023-02-02
Added new rule set for Keychron K6

Gsinti:
# Keychron K6 Useful Escape & Caps Lock Hyper Mod

This rule adds Escape hotkeys for Keychron K6 users to maintain muscle memory between the built-in MacBook keyboard and the K6.

It also adds the Hyper Key mod (⌃⌥⇧⌘) to Caps Lock when held for creating complex hotkeys that are easy to input. Caps Lock still works by simply pressing the key alone.

```
Some rules are borrowed from @sourabhv with slight tweaks:

    Left Shift + Esc acts as ~
    Control + Escape acts as Control + `

This matches the key positioning for users who also use the built-in keyboard.
```
_(Thanks to @sourabhv for the inspiration.)_

Full list of modifications:

- Unmodified Esc: Escape
- Left Shift + Esc: ~ (tilde)
- Right Shift + Esc: ` (grave)
- Control + Esc: Control + ` (iTerm hotkey)
- Command + Esc: ⌘ + ` (app window cycle)
- Caps Lock (held): ⌃⌥⇧⌘ (Hyper Key)
- Caps Lock (alone): Regular Caps Lock
- Hyper Key + Esc: Shift + Escape

### DIFF
--- a/public/extra_descriptions/keychron_k6_useful_escape_caps_lock.json.html
+++ b/public/extra_descriptions/keychron_k6_useful_escape_caps_lock.json.html
@@ -7,7 +7,7 @@
 </p>
 
 <p>
-  It also adds the Hyper Key mod (⌃⌥⇧⌘) to Caps Lock when held for creating complex hotkeys that are easy to input. Caps Lock still works by simply pressing the key once.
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

PR #1310, opened 2023-02-06
Add "Tab as a meta key" rules

joakim:
Hope others will find this as useful as I have.

It's especially useful for indenting/outdenting code in IDEs. But it's also a more intuitive way to move focus in any app.

    Tab+Left / Tab+Up    = Move focus back / Dedent
    Tab+Right / Tab+Down = Move focus forward / Indent

Treating the Tab key as a meta key makes more sense, IMO.

PS. Thanks for Karabiner!

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -157,6 +157,9 @@
         },
         {
           "path": "json/emacs_key_bindings_move_delete_select.json"
+        },
+        {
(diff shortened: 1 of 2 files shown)


Gsinti:
Looks great!

joakim:
Thanks :)

PR #1313, opened 2023-02-09
Exclude some apps from layer 3 rule

andre-krueger:
With these changes, Apple Terminal, Emacs and iTerm2 are excluded from the `Prevent all layer 3 keys from being treated as option key shortcut` and `Prevent problematic keys (?, /, #, =, and ')') from being treated as option key shortcut` rules, as with these rules enabled, extra characters would be added inside these apps when layer 3 is used.

Related: The included terminals in JetBrains IDEs still have the issue with added characters even with these changes (a tilde in that case), anyone know a fix for that?

### DIFF
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -14188,6 +14188,14 @@
                   "input_source_id": "^org\\.unknown\\.keylayout\\.DeutschADNW$"
                 }
               ]
+            },
+            {
+              "type": "frontmost_application_unless",
+              "bundle_identifiers": [
+                "com.apple.Terminal",
+                "org.gnu.Emacs",
+                "com.googlecode.iterm2"
+              ]
             }
           ]
         },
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

> Related: The included terminals in JetBrains IDEs still have the issue with added characters even with these changes (a tilde in that case), anyone know a fix for that?

Unfortunately, Karabiner-Elements doesn't support switching behavior inside the application.
In the past, the Accessibility API was used to provide features such as conditional behavior by window title.
But the side effects of using the Accessibility API were too great, such as preventing other apps (e.g., Adobe apps) from working properly, so there are no plans to support this in Karabiner-Elements at this time.

stapelberg:
I’m not 100% sure if the problem is related to this change, but when I import these new neo2 rules into karabiner, the rule “Prevent problematic keys from being treated as option key shortcut” no longer works in Google Docs in Google Chrome: I get “i” instead of “/” when I press mod3+i.

PR #1316, opened 2023-02-11
add pc style copy (shift+insert) as there already is paste in this file

mmoole:
there is also the file `public/json/PC-Style_Shortcuts_Copy_Ctrl_Insert.json` which includes only this modification. Maybe in order to prevent duplicates, this file could be deleted. 
Or one could also think it the other way and move the snippet containing paste to this file. 
But since this belongs to PC-style modifications nevertheless, I would suggest having both snippets included in the `pc_shortcuts.json` file.


tekezo:
Thanks!

PR #1335, opened 2023-03-06
add complex modifications for fn keys in VSCode

ayberkydn:
added a complex modification for fn keys to word as fn key is pressed in VSCode. shift+fn is also included. useful for debugging.


tekezo:
Thanks!

PR #1338, opened 2023-03-13
capslock as esc, without delay

skogsbaer:
All other configuration seem to have a slight delay before recognizing capslock as esc. This is because these other configurations have special behavior when capslock is pressed together with some other key.

But if you use vim, then esc must be recognized instantly.


tekezo:
Thanks!

PR #1339, opened 2023-03-14
fix volume down with down arrow

mngyuan:
I made this PR 3 years ago and don't use it much anymore because the keyboard I made it for I don't have anymore, but I noticed just now there's a bug.


tekezo:
Thanks!

PR #1349, opened 2023-03-21
Disable Neo complex modification in warp terminal

siboehm:
In the [Warp Terminal](https://www.warp.dev/) the complex modification "Prevent problematic keys from being treated as option key shortcuts" inserts extra spaces, when pressing e.g. "/".


tekezo:
Thanks!

PR #1354, opened 2023-03-29
Use PC Insert key to ScreenCapture to clipboard

unity1979:
Use PC Insert key to ScreenCapture to clipboard (Instead of screen capturing to file system.) Same as CMD+CTRL+SHIFT+4


tekezo:
Thanks!

PR #1367, opened 2023-04-18
Add Firefox Deveoper Edition to windows-on-mac shortcuts

bo-carey:
I noticed that for most of the public rules for use on browsers (such as Ctrl+L => Cmd+L) the bundle identifiers are missing `org.mozilla.firefoxdeveloperedition`. I'm sure many of the Karabiner Elements users are also using that app for and would appreciate it being included the relevant rules. Please feel free to add any that I missed, and I would really appreciate these rules being added for posterity! Thanks 😃


tekezo:
#1372

tekezo:
Thanks!
We also need to update src/lib/karabiner.rb to avoid regression of the generated json.
I added the changes to this commit in #1372.

PR #1370, opened 2023-04-19
Adding a new Citrix modifier to remove Fn key from functional keys

mpiotrowski:
When using Citrix with Windows machine, I often need to use F* keys. I would like to be able to use windows keys the same way while in Citrix without changing system settings


tekezo:
Thanks!

PR #1374, opened 2023-04-27
refactor(lshift_to_escape): Optimize logic

BerginJay:
In the original JSON, a short single press of `left_shift` generates down and up events for both `left_shift` and `escape`, which may cause trouble in some cases. The original description was not clear and actually described `left_shift` as `caps_lock`. 

The title and description are now more detailed. The old JSON generator file is removed. The logic is now changed to:
- short press alone: escape
- long press alone or with other keys: shift


tekezo:
Thank you for the correction to the appropriate content!

PR #1380, opened 2023-05-04
Treat Dictation as F5

gerrard00:
My external keyboard sends `consumer_key_code` set to `dictation` when hitting the F5 key even when the "Use F1, F2, etc. keys as standard function keys" is set to true. This is a rule that converts that to `F5` since I never use dictation and I'm sick of having to hold the `fn` button while using debuggers.


tekezo:
Thanks!

PR #1381, opened 2023-05-05
feat: Add LAlt option for open_terminal_window

wouter-n:
Extend `open_terminal_window` to have a shortcut for opening with the terminal with left Alt (like Ubuntu and some other Linux distros)


tekezo:
Thanks!

PR #1382, opened 2023-05-08
fixed ctrl_arrows_to_option_arrows so option+arrow acts like ctrl+arrow

jonathan-golorry:
For some reason the rule was remapping ctrl+arrow twice instead of properly swapping.


tekezo:
Thanks!

PR #1396, opened 2023-06-04
Clarified swapping ctrl+delete with option+delete and using ctrl+delete to forward delete

Jonathan-Landeed:
These should be easier to use. I think the two concepts got mixed up at some point.


tekezo (approved):
LGTM!

tekezo:
Thanks!

PR #1409, opened 2023-06-19
Add cmd shift left/right arrow rule

MPM10223:
Apologies in advance if I'm not contributing correctly.

I couldn't find a rule that would map `Cmd + Shift + L/R Arrows` to `Ctrl + Shift + L/R Arrows` (to mimic Windows functionality in text editors), so I added one.

Thanks for all your work on this project!


tekezo:
Thanks!

PR #1410, opened 2023-06-21
Fix German Keyboard Dead Keys: Backtick, Tilde, Circumflex

samuelraub:
This allows to type Backtick (`), Tilde (~) and Circumflex (^) in one keystroke on german keyboards.


tekezo:
Thanks!

PR #1411, opened 2023-06-27
Hold CAPS LOCK for ijkl arrow keys

Aldin38:
I use the (hold caps lock)  + ijkl for arrow keys on the Anne Pro 2 keyboard, so I wanted to add this feature here aswell


tekezo:
Thanks!

PR #1413, opened 2023-06-29
control + w,i,l,u -> delete left word,jump one word to the left/right…

sgst-dev:
- control + w -> control + delete: delete the left word
- control + i -> option + left_arrow: jump to the left of the left word
- control + l -> option + right_arrow: jump to the right of the right word
- control + u -> command + z: undo


tekezo:
Thanks!

PR #1415, opened 2023-06-30
Create rule "Backslash to forward delete"

KirillGoryakin:
Use Backslash (\\) to forward delete.
Use option + Backslash (\\) to delete forward a word
Use command + Backslash (\\) to delete forward a whole line

Use Fn + Backslash (\\) for it's default behavior


tekezo:
Thanks!

PR #1416, opened 2023-06-30
autoshift 2 - key down will be send immediately and then if key held there will be backspace then caps

Old-iddi:
This Autoshift 2 implementation will be useful for cases when key down event needed when key is down and not when it released (as is in original Autoshift). Actually, I have discomfort when don't see letter immediately after key down, so may be others have same issue.


tekezo:
Thanks!

PR #1419, opened 2023-07-03
Edit from https://github.com/kravchik left command

jordandraper:
Took the JKIL arrows provided by kravchik and modified them to use the left command instead of right


tekezo:
Thanks!

PR #1422, opened 2023-07-05
New complex mod, left shfit + caps lock = return

jordandraper:
Found this useful when needing a carriage return using my left hand


tekezo:
Thanks!

PR #1429, opened 2023-07-10
Create rule "Fn=Option+Space"

KirillGoryakin:
Fn = Option + Space.
For better changing input sources if you have Cmd + Space for Spotlight.
(Set Globe Key to do nothing and set Option + Space for \"Select the previous input source\" in Keyboard shortcuts)


tekezo:
Thanks!

PR #1436, opened 2023-07-22
Add SpaceFN plus

tshu-w:
This is an upgraded version of SpaceFN. The existing SpaceFN configuration unfavourably activates SpaceFN mode during rapid typing, rendering it unusability. The concept is drawn from [SpaceLauncher](https://spacelauncherapp.com), with two primary revisions:

1. Inhibit SpaceFN Mode for a duration of 0.2 seconds post the last keystroke: This adjustment allows for the quick insertion of a space following the completion of a word.
2. Ingore pre-defined keys within 0.2 seconds after space key down: This enhancement aims to prevent a letter from being activated before releasing the space bar during speedy typing.

I have been using these two [improvements](https://gist.github.com/tshu-w/82af1843798edaab9062e44cb261d25a) for 4 years, and they work well for me.


tekezo:
Thanks!

PR #1439, opened 2023-07-29
Add: Language Switch Rule for EN/JP/KR using Print Screen, Scroll Lock, and Application keys

oozoofrog:
this is modification from left_cmd-EN-right_cmd-CH-right_option-JA.json

thanx


tekezo:
Thanks!

PR #1440, opened 2023-07-31
Add eisuufn.json

mementorack:
EisuuFN is similar to SpaceFN and KanaFN, but more comfortable for Japanese keyboard users; the Eisuu key is in a thumb-friendly position on Japanese keyboards, and this setting also allows the FN key to be placed on the left side.


tekezo:
Thanks!

PR #1464, opened 2023-08-21
Swedish mx keys mappings for programmer

fredAlm:
Mappings for swedish programmer using MX keys keyboard on Mac (used to windows key mappings)


tekezo:
Thanks!

PR #1468, opened 2023-08-22
Personal set of hotkeys by mlewand

mlewand:
First of all, thanks for amazing project ❤️ It allowed me to do macros that I was only able to do with AutoHotkey on Windows.

This configuration aims to (ab)use caps lock as a modifier to have essential keys closer to the hot regions of keyboard.

~I added only source version as I believe public json will be built automatically based on the `.js` file~ [the CI](https://github.com/pqrs-org/KE-complex_modifications/actions/runs/5943427897/job/16118540348?pr=1468) has shown me that I also need to add a json build version :).

Thanks!


tekezo:
Thanks!

PR #1480, opened 2023-08-28
Updated personal mlewand settings

mlewand:
*   Added support for extra two key chords: `caps+shift+w→{, caps+shift+e→}`
*   Added mapping for `caps + tilde` to a print screen.


tekezo:
Thanks!

PR #1491, opened 2023-09-20
Add korean_swap_won_and_backtick.json

gwonhong:
There was similar rule: "Change Won (₩) to grave accent (`) in Korean layout" but there was no way to 'swap' between them. I wanted to type "₩" with option key down.

I made typing <\`> to be <left_option + right_option + \`>, and <option + \`> to be <\`> so that both works well even though applied simultaneously.


tekezo:
Thanks!

PR #1497, opened 2023-10-01
add a config for VSCode/Cursor & 1Password

jgautsch:
This config makes the `cmd + \` key command redirect to `cmd + F11` whenever VSCode or Cursor are the frontmost applications. 

There's already a similar community rule for _just_ VSCode, but I've started using Cursor and want the same thing.


tekezo:
Thanks!

PR #1506, opened 2023-10-11
Add Microsoft Sculpt Ergonomic Desktop ISO PT-pt device

edgarbs1998:
Add specific complex modifications for the Microsoft Sculpt Ergonomic Desktop device with ISO PT-pt keyboard layout.


tekezo:
Thanks!

PR #1515, opened 2023-10-18
Mimic PC Keyboard Shift+Insert with Shift+Delete on Mac

swinful:
New keybindings to Mimic PC Keyboard Shift+Insert with Shift+Delete on Mac.


tekezo:
Thanks!

PR #1529, opened 2023-11-03
remap: modify button2 to run after key up

willbchang:
This fixes ghost select in context menu. When you are holding right click button and move mouse in the context menu, then release the button, it will click the selected item. It could be annoying if someone right click, move the cursor right and release the button quickly. It will do the unexpected select and click.


tekezo:
Thanks!

PR #1531, opened 2023-11-05
Added Command Forward Delete Line

deafSpy:
Created a remap to forward delete the rest of the line using Cmd + Fn + Delete


tekezo:
Thanks!

PR #1537, opened 2023-11-10
Mouse3(button) as Command+Tab for Application Switcher

brianlamb:
Will cause mouse3 to Application Switch (Command-Tab) in macOS, quickly switch between two most recently used Open applications


tekezo:
Thanks!

PR #1539, opened 2023-11-11
'Caps→Ctrl/Esc if alone' and 'UK→US Mac keyboard'

mngyuan:
Two quality of life changes, one which changes the built in macbook keyboard on a UK mac to resemble the US mac keyboard, and one which allows capslock to function as ctrl when pressed in combination with other keys, or as escape when pressed alone.

I've been using these for 7 months now without incident and put them in a PR now as I'm setting up some new macs.


willbchang:
Thanks

PR #1542, opened 2023-11-17
Updated Windows to Mac Mappings

Steve-Morales:
New mapping, Updated Window mappings


Steve-Morales:
Updated Window mappings

tekezo:
Thanks!

PR #1543, opened 2023-11-20
Fix the bundleIdentifier of Microsoft Word

zalt50:
The identifier of Microsoft Word is incorrectly defined as ``microsoftExcel``.


tekezo:
Oh, I don't know why I didn't notice it before, thank you!

PR #1544, opened 2023-11-21
Add improved EISU and KANA functions

MANABU0SEC:
I've added improved EISU and KANA functions, which would be beneficial for Japanese users, especially those using JIS keyboards.


tekezo:
Thanks!

MANABU0SEC:
Thank you, too!

PR #1545, opened 2023-11-23
Add "Effective Utilization of Bottom Row" settings

MANABU0SEC:
I’ve added settings for effective utilization of bottom row. With these settings, we will be able to use Enter, Delete, and Shift with our thumb. Furthermore, if you are a JIS keyboard user, you will also be able to use Control and Option with your thumb in the home position.


tekezo:
Thanks!

MANABU0SEC:
Thanks a lot!

PR #1546, opened 2023-11-23
Tap left shift to open and right to close curly brace

ggerritsen:
Add Karabiner Elements rule to exchange shifts to curly braces


tekezo:
Thanks!

PR #1549, opened 2023-11-25
Add configs for keyboard with trackball such as Adesso WKB-3150UB

Minyus:
public/json/emulate_mouse_clicking_and_scrolling.json: emulate mouse clicking and scrolling so that the right-hand thumb can stay on the trackball

public/json/shortcut_delete_and_enter.json: shortcut Delete Back/Forward and Enter so that the far keys do not need


tekezo:
Thanks!

PR #1551, opened 2023-11-26
New rule for disabling external keyboard when MacBook is opened

Otter-man:
This is a rule for disabling any external keyboard when your MacBook is opened. This feature is particularly useful when you use your MacBook in clamshell mode, i.e., connected to an external display and keyboard/mouse, and wish to prevent the external keyboard from responding once the MacBook is opened, without the need to physically disconnect it.

The rule operates on a simple principle: it instructs your Mac to ignore any keystrokes from external keyboards when the internal variable macbook_clamshell_status in Karabiner is set to opened. Conversely, it disregards this rule when the variable is set to closed.

A specialized shell script, available on my [GitHub](https://github.com/Otter-man/karabiner-clamshell-monitor), is required for this rule to work.
This script, once installed, will run in the background of your system after every startup and will monitor clamshell status of your Mac via macOS event stream.

Instructions for this rule added as extra description html and also linked in the description of the rule in json file.


tekezo:
Wow, this is awesome! Thank you!

PR #1560, opened 2023-12-21
added old macbook numpad style

2364c66e322e44d99b0caeee94b8010c:
Added the old style numpad using fn+ "7890 uiop jkl; m./"


tekezo:
Thanks!

PR #1565, opened 2023-12-31
Launch Terminal with F4

jinyoungch0i:
Hi @tekezo! 👋🏼

I wanted to define a custom rule to switch out the default functionality of F4 to launch a new Terminal window. Thought it might be useful to others, so creating this PR to make this rule importable.

(Also minor but I noticed that the default macOS behaviour for `F6` (turn on `🌙 Do Not Disturb`) hasn't yet been mapped in [Karabiner Elements](https://github.com/pqrs-org/Karabiner-Elements/blob/10b1bc537dbd0befd9d0af315c9857477d8581b5/NEWS.md?plain=1#L231) and instead mapped to `f6`; wondering if that's so since the feature is relatively new/bit of a hassle to implement? 🙂)

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -5,6 +5,9 @@
       "id": "modifier-keys",
       "files": [
         {
+          "path": "json/f4-to-launch-terminal.json"
+        },
+        {
             "path": "json/cmd+fn+delete_forward_delete_line.json"
         },
         {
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

> bit of a hassle to implement

Yeah, we have been trying to operate Do Not Disturb for a long time, but we haven't found the key code and we haven't found a stable way to operate it programmatically, so it is now f6.

jinyoungch0i:
Got it, thanks for all that you do in any case, been a big fan of KE for some time. 

Happy new year!

PR #1567, opened 2024-01-04
Magic keyboard escape key sequence (CMD+.) to mac Escape

sming:
👋  First PR here so please educate me as necessary 🙏🏻

Basically, I find myself hitting Command_L+. all the time on my MBP ever since I bought a magic keyboard for my iPad, which - unsurprisingly - has Command_L+. (and Command_R+.) as a shortcut for Escape.

Update: I’ve just discovered - via https://www.idownloadblog.com/2020/05/15/escape-key-shortcut-tutorial/ - that CMD+. has been a MacOS shortcut for Escape forever 🤦🏻‍♂️ :

> While you cannot bring back a missing physical Escape key, you can replicate its functionality with a little help from a nifty universal keyboard shortcut, and here’s how:
> 
> To invoke Escape, press Command + period (.) on the keyboard.
> 
> This useful system-wide synonym for Escape works across iOS, iPadOS, and macOS. It actually dates back to classic Mac OS decades ago. With iPadOS 13.4, Apple’s brought the key combination to its tablets, providing not only feature parity in terms of the Escape command between the iPad and Mac but also a viable replacement for those who couldn’t live without it.
> 
> Note: The above keyboard shortcut may not work for all apps. To address this, you can remap the ESC key.
> 

However, when I checked this on Apple apps (App Store, Messages, Calendar), it did not perform as Escape, perhaps as noted in the final sentence of the above quote. Not on my MBP, anyway. But when I enabled my rule, it did perform as Escape, which was nice.

Hence I am unsure as to whether to proceed with this PR. I suppose the user always has the choice to not install/activate this complex mod.

Reviewers: what are your thoughts? Thanks.

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -1500,6 +1500,9 @@
         {
           "path": "json/macbook_disable_external_keyboard_on_open.json",
           "extra_description_path": "extra_descriptions/macbook_disable_external_keyboard_on_open.html"
+        },
+        {
+          "path": "json/command_l-and-period-to-escape.json"
         }
       ]
     },
--- a/public/json/command_l-and-period-to-escape.json
+++ b/public/json/command_l-and-period-to-escape.json
@@ -1,5 +1,5 @@
 {
-  "title": "Change Command_L + Period to Escape",
+  "title": "Change Command_L|Command_R + Period to Escape",
   "rules": [
     {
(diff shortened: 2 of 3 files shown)


tekezo:
Thanks!
Even the Safari address bar, command+. does not work, so actually changing to escape can be worthwhile.

PR #1568, opened 2024-01-04
feat: add caps to hyper or esc if alone modification

michnak:
This is inspired by: https://ke-complex-modifications.pqrs.org/#hyper_key - but with a slight modification and also info in description about how to still turn on Caps lock


tekezo:
Thanks!

PR #1569, opened 2024-01-05
fixed wrong mapping of 456 in fn_old_style_numpad rule

2364c66e322e44d99b0caeee94b8010c:
fixed wrong mapping of 456 in fn_old_style_numpad rule


tekezo:
Thanks!

PR #1572, opened 2024-01-08
Colemak Ru Keyboard Hotkeys Sync

lossy0011:
Remap keys to use Colemak keyboard hotkeys with Russian keyboard layout


tekezo:
Thanks!

PR #1576, opened 2024-01-12
Add devheedoo.json

devheedoo:
- Caps Lock to Esc if alone, to Left Ctrl if not alone
- Left Ctrl to Esc if alone
- Left Ctrl + hjkl to Arrow Keys
- Left Cmd + Esc to Left Cmd + `


tekezo:
Thanks!

PR #1579, opened 2024-01-18
Mapping JIKL to arrow keys, including many modifiers

annietma:
Overview of added key mappings. Note that J is interchangeable with I, K, or L, and ← is interchangeable with ↑, ↓, or →.

CleanShot 2024-01-17 at 23 27 49@2x


tekezo:
Thanks!

PR #1538, opened 2023-11-10
Neo2 layer3 fixes

andre-krueger:
~~These changes require that [this](https://git.neo-layout.org/neo/neo-layout/pulls/644) pull request is merged first.~~

PR #1594, opened 2024-02-25
Fix Neo2's layer 3 in Wezterm in its default configuration

Cu3PO42:
The rules for Neo 2 include optional rules that should prevent applications from treating the produced keys as shortcuts with Option held. However, these rules break in some terminal applications because they treat Option differently from normal applications (e.g. as ESC+key). Therefore, the rules are inactive on some terminal apps by default. I have simply added another terminal to the list that breaks with those rules active.

This will have merge conflicts with #1538, but they are trivially resolveable in whichever order they're merged in.

### DIFF
--- a/public/json/neo2.json
+++ b/public/json/neo2.json
@@ -6807,7 +6807,8 @@
                 "com.apple.Terminal",
                 "org.gnu.Emacs",
                 "com.googlecode.iterm2",
-                "dev.warp.Warp-Stable"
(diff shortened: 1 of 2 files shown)


tekezo:
@Cu3PO42 Thank you for updates. I merged #1538 just now, so could you resolve the conflict?

Cu3PO42:
Sure! I just did.

tekezo:
Thanks!

PR #1595, opened 2024-02-27
Add complex rules to allow 'ä' and 'Ä' via long-pressing 'a'

HealsCodes:
This PR adds a ruleset to solve https://github.com/pqrs-org/Karabiner-Elements/issues/3608


tekezo:
Thanks!

PR #1597, opened 2024-02-29
add caps_lock as an optional modifier

river:
Modified my personal rules. Added caps_lock as an optional modifier for every rule.


tekezo:
Thanks!

PR #1598, opened 2024-03-01
R_CMD to switch input languages, especially Korean

predict-woo:
This fixes the issue where R_CMD does not switch input languages when caps lock is on.
The original author was @pydemia


tekezo:
Thanks!

PR #1609, opened 2024-03-30
Add coder friendly F keys

jiarui-z:
For iTerm2, IntelliJ, and Visual Studio Code: Fn + F-keys trigger media controls, while F-keys alone function as standard F-keys.

For other apps, it behaves the other way around


tekezo:
Thanks!

PR #1611, opened 2024-04-04
Add a new caps lock mapping

brycekbargar:
I haven't found another modification that quite does all of these three things
* Caps Lock is Esc + Control
* Caps Lock functionality is restored as the Esc key
* Caps Lock + njkluio789 becomes a number pad without messing up other ctrl shortcuts.

The last is "achieved" by using right control instead of left control.

I've tested it by copying it into the complex mappings directory.


tekezo:
Thanks!

PR #1621, opened 2024-04-30
evernote-slashcommand-killer rule is added

AoVA:
I have added a key configuration for Evernote that automatically dismisses the slash command popup.
By quickly toggling the IME, this feature also works in Japanese input mode.
The original functionality of the slash key can now be accessed using Command-/.
The functionality previously assigned to Command-/ has been moved to Command-Shift-/.
Best regards,


tekezo:
Thanks!

PR #1623, opened 2024-05-10
Added my custom rule for change ctrl+h to backspace or delete

risto24:
Thank you for all your work. I've added what is written in the title.


tekezo:
Thanks!

PR #1633, opened 2024-05-28
feat: add wezterm to windows shortcuts's excluded bundles

vohoanglong0107:
# Summary

This PR adds [WezTerm](https://wezfurlong.org/wezterm/) to the list of excluded bundles for Windows shortcuts on Macs. WezTerm is a cross-platform terminal, similar to `kitty` and  `alacritty`. Therefore, Windows shortcuts like `Ctrl+C` and `Ctrl+A` should be disabled to prevent sending incorrect commands to WezTerm.


tekezo:
Thanks!

PR #1651, opened 2024-06-14
Add 'Double-tap play/pause to skip track' complex mod

yarinkaul:
This complex modification assigns the play/pause media button twice, so that it works as before with a single press of the button. If it is tapped twice, it will instead skip to the next track.

Tested on a Logitech MX Keys Mini connected to a 2021 MacBook Pro running Mac OS X 14.2.1.


tekezo:
Thanks!

PR #1681, opened 2024-06-21
fix: keypad_period is useful on Blender (graphical designer) Machenike k500w

filiperochalopes:
Update conditional modification for Blender can work


tekezo:
Thanks!

PR #1683, opened 2024-06-26
Ctrl for lesser-than and greater-than signs

perrotuerto:
As described in the title, this allows to use `<` & `>` with Ctrl. This is convenient for Latin American language in EN keyboard.


tekezo:
Thanks!

PR #1684, opened 2024-06-28
Disable Option-click minimizing windows

grugnog:
When pressing option and clicking between windows the active window is disabled. I use option to control widow management shortcuts, so often end up accidentally minimizing windows. Here is someone else running into this: https://discussions.apple.com/thread/252719638?sortBy=best

This will of course break any apps that have behavior attached to option-click, but I haven't hit any yet!


tekezo:
Thanks!

PR #1685, opened 2024-07-07
add new rule to toggle dark mode with Nightfall

bap2pecs:
https://github.com/r-thomson/Nightfall/ is a tool to let you easily toggle dark mode

Ctrl+Cmd+Opt+T is the shortcut


tekezo:
Thanks!

PR #1686, opened 2024-07-07
Update Tye-Tartarus-v2.json to fix "decrease magnification size" setting

quantum-designer:


quantum-designer:
fixed "decrease magnification size" button!

PR #1688, opened 2024-07-09
add new rule to toggle Dark Reader On/Off in Chrome/Brave

bap2pecs:
https://darkreader.org/ is a browser extension to enable dark mode on any websites

Shift+Option+D is the command to toggle it on and off

this is a companion rule with https://github.com/pqrs-org/KE-complex_modifications/pull/1685

lmk if you think it's better to merge those two and maybe rename the rule to dark_mode.json or similar


tekezo:
Thanks!

PR #1697, opened 2024-08-02
Add HU u-return modification

balintb:
Replaces `ű+return` with `return`. By doing so, attempts to reduce the number of occurrences where an `ű` char is emitted before return due to the proximity of these keys in the HU keyboard layout. A very common annoyance.


tekezo:
Thanks!

PR #1699, opened 2024-08-08
Add shift modifier for Home/end and FN+arrows presets

ClementPinard:
This PR modifies already existing presets.

This is designed to have a Linux like behaviour when pressing either home/end or FN + right or left

The presets already exist elsewhere, but now the shift modifier is added, because it can be used when you want e.g. to select all the text until the end of the line.


tekezo:
Thanks!

PR #1702, opened 2024-08-20
Use h/l to switch applications

shenyfg:
After command+tab, use h/l to choose applications, rather than the annoying arrow keys


tekezo:
Thanks!

PR #1703, opened 2024-08-22
add shift + keypad_plus modifier for search in vim -> * searches forward add shift to search backward - #

pavelz:
…ard add shift to search backward - #


tekezo:
Thanks!

PR #1718, opened 2024-09-27
Fix: Conditions were disabling Complex rule in some applications

gilitos92:
Some pre-determined conditions were disabling the Complex rule in some applications. These conditions are not necessary for the user of this rule as they vary from individual to individual, therefore, the best thing to do is to remove them while still letting the final user fill the rule if they wish to do so.


tekezo:
Thanks!

PR #1726, opened 2024-10-09
Readme add key tag make all

orkoden:
Adds a hint to use the html tag `<button>` for nicer buttons in html. 

Changed `make` to `make all` because just `make` didn't work for me the first time I ran it.


tekezo:
Thank you!

@orkoden 
I created a PR to use the `<kbd>` tag, which is specifically for representing key inputs. Could you please review it?
https://github.com/pqrs-org/KE-complex_modifications/pull/1729

PR #1727, opened 2024-10-10
Majestouch minila emulator

toshi-upper-field:
I have created a setting that emulates the FILCO Majestouch MINILA keyboard.
Please merge it.


orkoden:
You could also add an entry to groups.json and write an extra description as html. I will merge as is. Thank you for your contribution.

PR #1730, opened 2024-10-15
feat(Dvorak): Add The real progammer Dvorak layout based on ThePrimeagen configuartion

NoctuaCode:
This layout is a standard Dvorak layout, with tweaks on symbols and numbers row.


tekezo:
Thanks!

NoctuaCode:
My pleasure, thanks for the invite.

PR #1732, opened 2024-10-17
neo: Fix layer 2 lock

gifti258:
As of recently, pressing left shift and right shift only activates layer 2 when held down or holding down for about 2 seconds and then releasing.


NoctuaCode:
You may want to run `make run` in your local project before making your pull request.

gifti258:
Thanks, can't say I like how complicated the process is. Should have opened an issue instead.

PR #1735, opened 2024-10-18
feat(Dvorak): The real programmer Dvorak based on ThePrimeagen (Apple input devices only)

NoctuaCode:
I added a configuration for those who uses both laptop and programmable keyboards. So they don't have to switch profile or disable the complex modification.


tekezo:
Thanks!

PR #1739, opened 2024-10-26
Create rule to send Fn + Fx instead of Fx unless builtin keyboard

ansgarkroger:
From v15.2.0 of Karabiner Element the F1-F12 key events sent by Karabiner-Elements, macOS will convert them once again into media keys (e.g., sending F10 will trigger mute).
This rule will send the F1-F12 keys as if you were pressing the fn key on an external keyboard.


tekezo:
Thanks!

PR #1753, opened 2024-12-05
Fix PC-Style Lock Screen

FloopCZ:
The Control+Shift+Power does not work properly on my macOS Sonoma 14.5 with only PC-Style Shortcuts enabled in Karabiner 14.13.0. Using Control+Command+q fixes the issue for me and locks the screen properly.


tekezo:
Thanks!

PR #1756, opened 2024-12-07
updated the Bundle Identifier for Alacritty to the latest version

masisz:
I noticed that the Bundle Identifier for Alacritty was outdated, so I decided to create a pull request to address it. Please review, and if there are no issues, kindly merge it.


tekezo:
Thanks!

PR #1757, opened 2024-12-10
Add New Bulgarian phonetic keyboard layout

dilomo:
Add the new layout for the keyboard of Macs to support Bulgarian phonetic that is available only on Windows devices but is much more common than the provide one by apple.


tekezo:
Thanks!

PR #1761, opened 2024-12-21
Fix description in SIGKILL frontmost application

kidonng:
Learned a shortcut from this rule, but as can been seen in the following screenshot the native shortcut for force quitting apps is Shift+_Option_+Cmd+Esc, not Ctrl:

CleanShot 2024-12-21 at 19 26 53@2x


tekezo:
Thanks!

PR #1765, opened 2024-12-28
feat: add ghostty to list of terminal bundle identifiers

jimeh:
[Ghostty](https://ghostty.org/) is a new terminal emulator by [mitchellh](https://mitchellh.com/).


tekezo:
Thank you!
The js file is processed by the Duktape engine, which does not support backticks (`). Could you please change them to single quotes (’) instead?

jimeh:
🤦‍♂️...they were supposed to be single-quotes originally. I guess that's what I get for using a quick toggle-quotes command in Emacs while half-asleep... lol

tekezo:
Thanks!

PR #1770, opened 2025-01-10
New rule to swap right_command with right_option

mircobabini:
This is particuraly useful if you use a windows-alike external keyboard on mac, or a win/mac keyboard like:

image
This is the Logitech MX Keys S. 

So, the right OPTION key is too far from the command, which is not familier with mac default keyboard. So it's pretty hard for  a mac user to hit right_command+ò to insert the @ (at symbol).

This basically switched the right_command to the right_option, and reverse.

### DIFF
--- a/public/json/swap_right_command_and_right_option.json  
+++ b/public/json/swap_right_command_and_right_option.json  
@@ -1,0 +1,35 @@
+{
+  "title": "Swap Right Command and Right Option globally",
+  "maintainers": [
+    "mircobabini"
+  ],
+  "rules": [
+    {
+      "description": "Swap Right Command and Right Option for all keyboards.",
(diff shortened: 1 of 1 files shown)


tekezo:
@mircobabini Thank you for PR!
The file `swap_right_command_and_right_option.json  ` that you added seems to have some trailing spaces in its filename. Please try renaming the file to remove the spaces.

mircobabini:
> @mircobabini Thank you for PR! The file `swap_right_command_and_right_option.json ` that you added seems to have some trailing spaces in its filename. Please try renaming the file to remove the spaces.

Weird. Fixed!

tekezo:
Thanks!

PR #1774, opened 2025-02-03
Remove myself as maintainer

harmtemolder:
I don't actually use Karabiner Elements any longer, and so cannot maintain any of the mods I wrote. This PR removes myself as maintainer


tekezo:
Thanks!

PR #1785, opened 2025-02-17
Add Caps Lock to debug in various IDEs

logicxd:
## Description

Map Caps Lock to Debugger Commands for IDEs such as Xcode, Intellij, and Visual Studio Code (VSCode). Current support: Step Over, Step In, Step Out, Continue

## Tested

1. On all 3 IDEs that the commands work
2. Checked out changes with `make preview-server` and applied to Karabiner


duleorlovic (approved):
Thanks!

PR #1797, opened 2025-03-24
feat(*): add swap ctrl shift with ctrl space func

tuanpham1406:
This function remaps the default keyboard shortcut for switching input sources on macOS, making it more convenient for Vietnamese users familiar with input methods like VietKey. It swaps the Ctrl+Shift combination with Ctrl+Space to streamline language switching and enhance typing efficiency.


tekezo:
Thanks!

PR #1798, opened 2025-03-26
add toggling Kotoeri punctual style with control period

TetsuakiBaba:
This is a configuration file that allows you to easily switch the punctuation style ([，．] / [、。]) in Kotoeri using the Control + Period shortcut. Since Apple does not officially provide an API for this, it runs a somewhat tricky script. It has been tested on macOS Sequoia 15.1.1, but it may not work properly on older versions.


tekezo:
Thanks!

PR #1809, opened 2025-04-30
feat: add korean won to backtick with shortcut preserved

sts07142:
Similar rules exist in #375  #1491 .
However, these existing rules have implications for shorcut using backticks.

Therefore, this rule converts only a standalone press of the won (₩) key to a backtick (`).
If the won (₩)  key is pressed together with any other modifier key, it behaves normally and does not interfere with existing shortcuts.


tekezo:
Thanks!

PR #1811, opened 2025-05-06
Add a rule to prevent accidental close window(⌘⇧W) in Google Chrome for ei-kana-cmd

darquro:
When the left and right Command keys are configured as language switch keys, it’s possible to accidentally close a Google Chrome window with ⌘+Shift+W.
To prevent this, I added a rule to disable the shortcut unless it’s pressed twice in a row.


tekezo:
Thanks!

PR #1812, opened 2025-05-08
HHKB arrows using Right Shift

nagnebro:
Although there's already a similar PR (#911), I need this option because the Right Shift key is positioned similarly to the actual Fn key on the HHKB.

### Key Mapping List
<kbd>⇧Right-Shift</kbd> + <kbd>;</kbd> = ←
<kbd>⇧Right-Shift</kbd> + <kbd>'</kbd> = →
<kbd>⇧Right-Shift</kbd> + <kbd>[</kbd> = ↑
<kbd>⇧Right-Shift</kbd> + <kbd>/</kbd> = ↓


tekezo:
Thanks!

PR #1813, opened 2025-05-15
Add rule "Swap Command and Option for all non-Apple keyboards"

FelixDuo:
Provide a config for those who has apple magic keyboard and other keyboards, periodically switch among those keyboards. Swap command and option keys on the left, and change the right option key to comand key as usually there is no right windows/command key for windows keyboard.


tekezo:
Thanks!

PR #1815, opened 2025-05-26
Change ANSI layout's keymap to support Dell JIS keyboard's symbol

enceka:
Tested fine with the travel keyboard of Dell 5290 2in1
Just modified the  symbol key (without ¥)


enceka:
keyboard like that
IMG_20250527_000416

PR #1832, opened 2025-08-01
Add custom rules for Keychron K6 Spanish layout

bguzmanm:
This PR adds a new set of complex modifications for the Keychron K6 keyboard configured for Spanish (Latin American) layout.

The configuration includes:

- fn + ' → acute accent dead key (´)
- fn + n → ñ

These mappings help users type accented characters more easily on macOS when using a Spanish layout with the Keychron K6.


tekezo:
Thanks!

PR #1833, opened 2025-08-03
feat: Add intuitive Spanish accent mapping for Keychron K6

bguzmanm:
This PR introduces a new complex modification designed to significantly improve the Spanish typing experience on the Keychron K6 keyboard. It provides a more ergonomic and intuitive way to type accented characters and the letter ñ, which is especially helpful for users accustomed to Latin American (ISO) keyboard layouts.

Instead of using the default macOS key combinations, this feature maps the [ key as a dedicated accent trigger. This allows for a faster and more natural workflow by mimicking the muscle memory of standard Spanish keyboards.

Key Functionality:

- Intuitive Accents: Press [ followed by a vowel to instantly get the accented character (e.g., [ + a = á).
- Easy `ñ`: Simply press [ followed by n to type ñ.
- Seamless Capitalization: The mapping works with the Shift key to produce uppercase accented letters (Á) and Ñ.
- This enhancement aims to make the Keychron K6 a more viable and comfortable option for Spanish-speaking users, reducing friction and improving typing speed.

### DIFF
--- a/public/extra_descriptions/keychron-k6-es-latam.json.html
+++ b/public/extra_descriptions/keychron-k6-es-latam.json.html
@@ -1,0 +1,40 @@
+<link rel="stylesheet" href="../../vendor/css/bootstrap.min.css" />
+
+<h2>Keychron K6 for es-LATAM users</h2>
+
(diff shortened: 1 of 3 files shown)


tekezo:
Thanks!

PR #1837, opened 2025-08-15
Add support for Naginata key layout

kojiOnHill:
This is to add support for Naginata key layout for Japanese input v.16 ([release note in Japanese](http://oookaworks.seesaa.net/article/509198141.html#gsc.tab=0)).


.gitignore
```diff
@@ -4,3 +4,4 @@
 *.tmp
 /public/build/
 /dist/
+go.sh
```
tekezo, line 7: For personal files that aren't part of the repository, it's better to ignore them via .git/info/exclude.

```suggestion
```

tekezo (approved):

PR #1841, opened 2025-08-18
Version 0.2

zakkoo:
My newest version has a developer mode and a description.


zakkoo:
@tekezo More is comming and and it will be refined (especially the description) since a lot of my coworkers will use it.

PR #1850, opened 2025-09-03
navigation layer and combos for 54 buttons keyboard

bskaplou:
Modification creates navigation layer accessible through CapsLock hold and adds combos.


tekezo:
Thanks!

PR #1852, opened 2025-09-08
Create f18.json

samknutson:
Add a short and simple example of remapping a single function key or other key to a key with modifiers.  This example is the default for Capture with SnagIt so mapping it to an extra function key on an external keyboard.  This is a better example to start with for many purposes than the complicated ones for other Hyper keys.

### DIFF
--- a/public/json/f18.json
+++ b/public/json/f18.json
@@ -1,0 +1,18 @@
+{
+    "description": "Remap F18 to Control+Shift+P useful for SnagIt or example simple remap of F13-Fnn",
+    "manipulators": [
+        {
+            "from": {
+                "key_code": "f18",
+                "modifiers": { "optional": ["any"] }
+            },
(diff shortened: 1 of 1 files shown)


tekezo:
Sorry the README was unclear. I'll fix it later.

For JSON distributed in this repository, you need to wrap it in a top-level object with a separate title and a rules array, like this to provide multiple rules in one file:

```
{
  "title": "your configuration title",
  "rules": [
      rule1,
      rule2
  ]
}
```

The entire example is here:
https://github.com/pqrs-org/KE-complex_modifications/blob/main/public/json/example_device.json

tekezo:
I've updated the document.
https://github.com/pqrs-org/KE-complex_modifications/?tab=readme-ov-file#json-file-format-in-this-repository

PR #1853, opened 2025-09-09
Add rules for just the punctuation differences on a German Windows layout

AplusKminus:
Other existing rulesets either:
- include modifier changes
- allow `modifiers: { optional: ['any'] }` (which is unnecessarily permissive and potentially breaks complex key-combinations)
- are built to be used with US-layout set in the OS


tekezo:
Thanks!

PR #1863, opened 2025-09-17
Add Corsair MARK 75 to Apple UK keyboard layout mapping

flipdazed:
This PR adds a new complex modification for the Corsair MARK 75 keyboard to properly work with macOS using the UK Apple keyboard layout.

## What does this modification do?
This modification makes the Corsair MARK 75 keyboard (VID: 6940, PID: 11012) behave like a native Apple UK keyboard by:

### Key Remappings:
- **Grave/Tilde key** (top-left, left of 1): Maps to the Apple UK dead accent key position
- **ISO key** (between Left Shift & Z): Properly outputs backslash `\` and pipe `|`
- **Hash/Tilde key** (left of Enter): 
  - Normal press → `#` (via Option+3)
  - Shift press → `~` (via Shift+non_us_backslash)
- **Quote key** (right of semicolon): Swaps with Shift+2 to fix quotation marks:
  - `'` → `'` (apostrophe)
  - Shift+`'` → `@`
  - Shift+2 → `"` (quotation marks)
- **Euro symbol**: Option+4 → `€` (remapped to Option+2 as per Apple UK layout)
- **Cmd ↔ Ctrl swap**: Swaps Command and Control keys for PC-style keyboard muscle memory

### Why is this needed?
The Corsair MARK 75, while having a UK physical layout, sends different key codes than what macOS expects for a UK keyboard. This causes incorrect characters to appear, particularly for symbols like `@`, `"`, `#`, `~`, and the Euro symbol. This modification ensures all keys produce the expected UK characters.

## Testing
- Tested on macOS with Corsair MARK 75 keyboard
- All UK layout special characters work correctly
- Modifier key swap (Cmd/Ctrl) functions as expected
- Device-specific conditions ensure it only affects the Corsair MARK 75

## Files changed
- `public/json/uk_apple_to_uk_corsair_makr.json` - New complex modification file
- `public/groups.json` - Added entry in "Device Specific" section


tekezo:
Thanks!

PR #1870, opened 2025-09-23
Fix incorect keybinds

michaelbull:
The windows-pc-colemark-jetbrains.json file has multiple erroneous keybinds.
    
The first describes changing the End key to instead perform as Control+E, however its implementation erroneously mapped to "k" and not "e".
    
The second describes changing Reload behaviour of Command+R to Control+R and changing F5 to Command+R, however its implementation erroneously used "s" for both of these instead of "r".


tekezo:
Thanks!

PR #1941, opened 2026-03-15
Add Noted layout support to Neo2

v-ji:
This PR adds support for the Noted keyboard layout ([introduction](https://dariogoetz.github.io/noted-layout/), [documentation (German)](https://www.neo-layout.org/Layouts/noted/)) to the existing Neo2 rule set.

I opened a [companion PR on the Neo repository](https://git.neo-layout.org/neo/neo-layout/pulls/677) to add the macOS keyboard layout file (.keylayout) for Noted, as the two depend on each other. I’m opening this one as a draft and warmly invite testing and feedback from the Neo community.

I noticed that the source file `src/json/neo2.json.js` wasn’t properly formatted based on the project’s Prettier configuration, so I did that first.

The actual changes:

- Added DeutschNoted to the `isLayoutActive` input source condition so that the Neo2 rules apply when Noted is active.
- Noted places its right Mod3 key further inward, on the quote key rather than backslash (where all other Neo-family layouts have it). To handle this difference, the `neo2Modifiers` helper was split into two:
  - `neo2Mod3Key` now takes an input-source condition so it can be applied separately for backslash (all layouts except Noted) and quote (Noted only).
  - `neo2Mod4` extracted from the former combined helper, shared across all layouts unchanged.
- Two new conditions (`isRightMod3Backslash`, `isRightMod3Quote`) gate each Mod3 binding to the correct layouts.

It might be worthwile to also add the [Mine layout](https://www.neo-layout.org/Layouts/mine/) to the `isRightMod3Quote` input source array, as it shares the same Mod3 key location with Noted. There is currently no macOS .keylayout file for it, but if someone wants to create it, they would not need an additional PR here.


tekezo:
I'm closing this for now due to inactivity.

PR #1943, opened 2026-04-08
Fix the pipe char for Swedish Logitech keyboards

fredAlm:
Add another modification to fix the pipe char for the Swedish setup


tekezo:
Thanks!

PR #1945, opened 2026-04-15
Add Gmail send-shortcut fix rule (Cmd+Shift+D to Cmd+Return)

DanMelbourne:
Remaps Cmd+Shift+D to Cmd+Return when a web browser is frontmost so the Mac-standard "Send email" shortcut sends the message in Gmail instead of triggering Gmail's built-in "Discard draft" shortcut. (Unfortunately Gmail does not expose the discard shortcut in its keyboard-shortcut settings, so it cannot be rebound from within Gmail.)

Supported browsers: Google Chrome (including PWAs), Safari, Arc, Firefox, Microsoft Edge, Brave.

Also adds an extra_description HTML with context and optional instructions for narrowing the remap to a Gmail PWA only.


tekezo:
Thanks!

PR #1949, opened 2026-04-20
Add right_control hold for Doubao voice input rule

Molunerfinn:
## Summary

- Add a new complex modification rule: hold `right_control` to switch to Doubao IME and activate `fn` key for voice input; on release, switch back to WeChat IME with Chinese input mode restored.
- Register the rule under the "International (Language Specific)" group in `groups.json`.

## Details

**On press:**
1. Switch input source to Doubao IME (`com.bytedance.inputmethod.doubaoime.pinyin`)
2. Wait 50ms for the input source to settle
3. Activate `fn` key (triggers Doubao's voice input)

**On release:**
1. Wait 1 second for pending voice input to commit
2. Switch input source back to WeChat IME (`com.tencent.inputmethod.wetype.pinyin`)
3. Wait 50ms, then press `Shift` to restore Chinese input mode

## Test plan

- [x] `make all` passes with no errors
- [ ] Rule imported and tested in Karabiner-Elements

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -1121,6 +1121,9 @@
           "path": "json/left_option_to_chinese.json"
         },
         {
+          "path": "json/right-ctrl-doubao-voice-input.json"
+        },
+        {
           "path": "json/japanese.json"
         },
         {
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #1950, opened 2026-04-26
Add Redragon Lakshmi 60% ABNT2 physical -> macOS US International

salesdbs:
## Summary

Keyboard mapping for the **Redragon Lakshmi 60%** with ABNT2 physical layout used with macOS **US International** input source (device VID 9610 / PID 42).

### What it fixes

- **Modifier swap**: Left/Right Alt <-> Command, Left Win <-> Option
- **ESC multifunctional**: solo = Escape | Shift+ESC = `"` | Option+ESC = `'`
- **Cedilha**: semicolon key -> c-cedilla (dead key + c sequence)
- **Dead keys**: Trema, Til, Circumflex, Grave accent
- **Brackets/Braces**: fixes inversion caused by ABNT2 -> US layout mismatch (`[`, `]`, `{`, `}`)
- **Punctuation**: backslash, pipe, colon, semicolon, slash, question mark

Rules are split into 4 logical groups so users can enable only what they need.

### DIFF
--- a/public/json/redragon_lakshmi_abnt2_us_intl.json
+++ b/public/json/redragon_lakshmi_abnt2_us_intl.json
@@ -1,0 +1,158 @@
+{
+  "title": "Redragon Lakshmi 60% ABNT2 physical → macOS US International (rev 1)",
+  "maintainers": ["salesdbs"],
+  "rules": [
+    {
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

PR #1951, opened 2026-04-26
Enable shift for Layer 4 in neo_layouts_based_on_abc_extended

luca-schlecker:
Using shift with the layer 4 movement keys does not select text. This change addresses this shortcoming.


tekezo:
Thanks!

PR #1952, opened 2026-04-30
Add Smart Caps Lock rule

vbwx:
This rule makes Caps Lock behave like Control when held, like Escape when pressed, and preserves its Caps Lock functionality when pressed in conjunction with Shift.


tekezo:
Thanks!

PR #1955, opened 2026-05-09
logitech mk235 hun

sanyacode:
Adding support for Hungarian MK235 for most frequently used keys.

Tested on local server successfully


tekezo:
Thanks!

PR #1957, opened 2026-05-15
Add Belgian AZERTY Windows AltGr mapping

davidvuy:
Adds a community rule for Belgian AZERTY PC keyboards used with the built-in macOS Belgian layout.\n\nThe rule keeps normal Command shortcuts available while mapping Right Command as Windows AltGr for common Belgian Windows symbols such as @, #, brackets, braces, backslash, pipe, tilde, acute accent, and euro. It also normalizes the PC ISO < > key beside Shift when macOS treats it as the Apple @/# key.\n\nValidation:\n- make all\n- karabiner_cli --lint-complex-modifications public/json/belgian_azerty_pc_windows_altgr.json

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -1144,6 +1144,9 @@
           "extra_description_path": "extra_descriptions/ei-kana-cmd.json.html"
         },
         {
+          "path": "json/belgian_azerty_pc_windows_altgr.json"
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #1960, opened 2026-05-23
shift switch input method for Mac default

albert-get:
It's customary for Chinese users to switch input methods with Shift.
Single press Shift switches input language, long press keeps original Shift function.
Verified and running correctly.


tekezo:
Thanks!

PR #1961, opened 2026-05-24
Add DoubaoIME voice input switch rule

Coldin04:
## Summary
- Add a DoubaoIME voice input switch rule triggered by double tapping right_command.
- Keep right_command available as a normal modifier unless double tapped.
- Add an extra description with customization notes and the original article link.

## Validation
- Ran make all
- Tested locally in Karabiner-Elements


tekezo:
Thanks!

PR #1972, opened 2026-06-01
Add windows style IME switcher rule

Elias-Lee-SC:
## Summary
- Add windows_style_IME_switcher complex modification
- Support Windows-style input source switching with Control+Shift tap and Control+Space
- Add the rule to International (Language Specific)

## Validation
- make all

PR #1973, opened 2026-06-02
Fix Control+Space exact matching in windows_style_IME_switcher

Elias-Lee-SC:
Follow-up to #1972.

## Summary
- Fix Control+Space matching in windows_style_IME_switcher to require the exact Control+Space shortcut.
- Remove optional "any" modifiers from both Control+Space manipulators.
- Keep the existing English -> next input source and non-English -> English behavior unchanged.
- Allow shortcuts with extra modifiers, such as Control+Command+Space, to pass through to macOS or other apps.

## Validation
- make all

### DIFF
--- a/public/json/windows_style_IME_switcher.json
+++ b/public/json/windows_style_IME_switcher.json
@@ -90,9 +90,6 @@
             "modifiers": {
               "mandatory": [
                 "control"
-              ],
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

PR #1974, opened 2026-06-05
Add extra description for windows_style_IME_switcher

Elias-Lee-SC:
Follow-up to #1973.

## Summary
- Add an extra description for windows_style_IME_switcher on the import site.
- Document the required macOS Input Sources shortcut settings before enabling the rule.
- Link the description from public/groups.json.

## Validation
- make all


tekezo:
Thanks!

PR #1978, opened 2026-06-24
Add "Mouse as a trackpad replacement" rules (Spaces on side buttons, Mission Control / App Exposé on middle button)

TyndaleLym:
This adds a small package that turns extra mouse buttons into trackpad-style navigation on macOS, so a regular mouse can cover the gestures usually done on a trackpad.

## Rules (each can be enabled independently)

1. **Side buttons switch Spaces** — `button4` → next desktop (`^→`), `button5` → previous desktop (`^←`).
2. **Middle button (`button3`)** — single click (< 300 ms) opens Mission Control / all windows (`^↑`); long press (≥ 300 ms) opens App Exposé / windows of the front app (`^↓`).

## Notes

- Validated locally with `make all`.
- Registered in the **OS Functionality** group in `public/groups.json`, next to the existing mouse desktop-switcher rules.
- The single-click vs. long-press middle-button behavior isn't covered by the existing `mouse3_app_switcher` / `mouse_buttons_4_5_desktop_switch` rules, so this packages all three buttons into one cohesive trackpad-replacement set.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/public/groups.json
+++ b/public/groups.json
@@ -154,6 +154,9 @@
           "extra_description_path": "extra_descriptions/mouse3_app_switcher.json.html"
         },
         {
+          "path": "json/mouse_trackpad_replacement.json"
+        },
+        {
           "path": "json/r_cmd_to_l_ctrl_and_space.json",
           "extra_description_path": "extra_descriptions/r_cmd_to_l_ctrl_and_space.json.html"
         },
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #1979, opened 2026-06-25
Add ABNT2 fix for swapped pipe/backslash and grave/tilde keys

mateuslecchi:
On external ABNT2 (Brazilian Portuguese) keyboards on macOS, the pipe/backslash key (non_us_backslash) and the grave/tilde key (grave_accent_and_tilde) are swapped. This rule swaps them back so they type the expected characters.

Adds the JSON modification, a bilingual (EN/PT) extra description page, and a groups.json entry under "International (Language Specific)".


tekezo:
Thanks!

PR #1987, opened 2026-08-01
feat: Support for the Keychron v6 special keys mapping

jucrramirez:
- The rules map the Keychron V6 Max special keys to it's "natural" behaviour in Linux/windows


tekezo:
Thanks!

PR #1989, opened 2026-08-06
Add Windows-style shortcuts: browser and Finder fixes

renovys:
Three small rules that fix conflicts I ran into while using PC-style shortcut sets on macOS.

### 1. Address bar: `⌘D` → `⌘L`, in browsers only (Windows `Alt+D`)

Existing PC-style sets map `Alt+D` globally. Combined with an `Alt` → `Command` modifier mapping that means `⌘D` is swallowed everywhere, so Finder's Duplicate and the browser's Add bookmark both stop working. Scoping the rule to browsers keeps `⌘D` intact elsewhere.

### 2. Bookmark: `control+D` → `⌘D`, in browsers (Windows `Ctrl+D`)

Restores bookmarking with the Windows shortcut. It emits `right_command`, while rule 1 matches `left_command` only, so rule 1 cannot re-capture rule 2's output. That makes the pair safe regardless of the order the two rules are enabled in.

### 3. Finder: `⌘F4` closes the window (Windows `Alt+F4`)

Finder has no Quit, so the usual `Alt+F4` → `⌘Q` mapping does nothing there. This closes the front window instead.

### Validation

- `make rebuild` — no regenerated diff
- `karabiner_cli --lint-complex-modifications` — ok
- In daily use on Karabiner-Elements, with both a Bluetooth and a 2.4 GHz keyboard

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014g3nfJvM7zQnQCV9tWabnt

### DIFF
--- a/public/json/windows_style_browser_and_finder_fixes.json
+++ b/public/json/windows_style_browser_and_finder_fixes.json
@@ -1,0 +1,105 @@
+{
+    "title": "Windows-style shortcuts: browser and Finder fixes (rev 1)",
+    "maintainers": ["renovys"],
+    "rules": [
+        {
+            "description": "Address bar: ⌘D to ⌘L, in browsers only (Windows Alt+D). Enable this BEFORE the Ctrl+D bookmark rule below. (rev 1)",
+            "manipulators": [
+                {
+                    "type": "basic",
+                    "conditions": [
+                        {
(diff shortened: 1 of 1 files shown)


tekezo:
Thanks!

PR #2007, opened 2026-09-05
Function keys remap Apple Silicon

marcopelegrini:
In Apple-Silicon/newer-keyboard the physical F-row reports a plain Keyboard-page usage code (e.g. f7) for the raw HID event regardless of whether Fn is held — the "media key vs. function key" distinction

The actual "still needs fn" symptom comes from Java AWT on macOS, not the hardware: when fnState=0, AWT only sets the Function modifier flag on a KeyEvent if the OS says Fn was involved, and JetBrains IDEs bind actions like fn F7 requiring that flag. This has been AWT/macOS behavior for years, independent of chip architecture — it just becomes more noticeable on newer keyboards (especially external ones, Studio Display keyboard, etc.) where the dual-role F-key handling is implemented slightly differently at the driver level than older pre-T2 Macs.

Tested on M4 with Jetbrains, VSCode and other apps


tekezo:
Thanks!

PR #2008, opened 2026-09-10
Fix Spotlight shortcuts in Windows Keyboard for Mac

Fuzzy-and-Fluffy:
Windows-key taps and Win+R/S in the Windows Keyboard for Mac community rule use `open_application` for `com.apple.Spotlight`. On our macOS Tahoe 26.2 machine, all three shortcuts produced no visible search UI.

This changes the four outputs (left/right tap, Win+R, Win+S) to `apple_vendor_keyboard_key_code: "spotlight"` with repeat disabled, matching Karabiner's native Apple Spotlight key action. Tap timing, lazy modifiers, remote exclusions, and Win+Space remain unchanged. The generator and generated JSON are updated together.

Validation: `make all` passed; the generated JSON also passed the installed Karabiner-Elements 16.1.0 CLI lint. Normalizing the four output objects back to the old action makes the generated JSON identical to the upstream version. A physical Windows keyboard on macOS 26.2 confirmed that a quick Windows-key tap, Win+R, and Win+S now visibly open Spotlight after applying the same output changes to the device-scoped companion profile. The initial failure was on Karabiner 16.1.0, while the post-retest environment reported 16.3.0; this was not a controlled comparison on one Karabiner version. This was a device-scoped local test, not a separate all-keyboards community-profile smoke test.

Related report: https://github.com/Fuzzy-and-Fluffy/windows-keyboard-for-mac/issues/7

### DIFF
--- a/public/json/windows_keyboard_for_mac.json
+++ b/public/json/windows_keyboard_for_mac.json
@@ -3030,11 +3030,8 @@
           },
           "to": [
             {
-              "software_function": {
-                "open_application": {
-                  "bundle_identifier": "com.apple.Spotlight"
-                }
-              }
+              "apple_vendor_keyboard_key_code": "spotlight",
+              "repeat": false
             }
           ],
           "conditions": [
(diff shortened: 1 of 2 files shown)


tekezo:
Thanks!

PR #2010, opened 2026-09-14
dvorak_layout: allow fn as an optional modifier

cxa:
## Problem

While `fn` is held, the Dvorak remap stops working and keys fall through to QWERTY.

`from.modifiers.optional` lists `caps_lock` and the command/control/option modifiers, but not `fn`. Karabiner-Elements matches a manipulator only when every pressed modifier appears in `mandatory` or `optional`, so holding `fn` makes all 138 manipulators miss and the original key code is posted unchanged.

This affects every remapped key — on a built-in Apple keyboard, <kbd>fn</kbd>+<kbd>k</kbd> posts `k` instead of `t`.

## Change

Add `fn` to `optionalModifiers` in `src/json/dvorak_layout.json.js`, and regenerate `public/json/dvorak_layout.json` with `make all` (138 × `"fn"`, nothing else).

`fn` stays optional rather than mandatory, so it is passed through rather than consumed.

One consequence worth calling out: macOS Globe-key shortcuts that are bound to letters (<kbd>fn</kbd>+<kbd>E</kbd> emoji picker, <kbd>fn</kbd>+<kbd>Q</kbd> Quick Note, <kbd>fn</kbd>+<kbd>C</kbd> Control Center, …) now follow the Dvorak layout — physical <kbd>D</kbd> posts <kbd>fn</kbd>+<kbd>e</kbd> and opens the emoji picker. That is consistent with how <kbd>⌘</kbd>/<kbd>⌥</kbd>/<kbd>⌃</kbd> shortcuts already behave under this rule, since those modifiers are already in `optional`, and with `src/json/real_prog_dvorak.json.js`, which uses `optional: ['any']` and so never had this problem.

Keys this rule does not remap (function keys, arrows, delete) are untouched either way.

## Testing

Applied the same one-word change to a local copy of this rule (Karabiner-Elements 16.3.0, built-in Apple keyboard) and confirmed <kbd>fn</kbd>+<kbd>k</kbd> posts `t`; before the change it posted `k`.

`make rebuild` leaves the tree clean apart from the two files in this PR.


tekezo:
Thanks!

PR #2011, opened 2026-09-18
Add Fn Lock toggle for the top row

anguswww:
## What this does

Adds a complex modification that toggles Fn Lock with Fn+Esc. While enabled, F1–F12 are sent with Fn so the top row produces function keys; while disabled, the normal media-key behavior is preserved. The toggle displays a short On/Off notification.

## Validation

- `make all`
- `karabiner_cli --lint-complex-modifications public/json/fn_lock.json`

Includes the generated JSON, OS Functionality category entry, and usage notes.


tekezo:
Thanks!

